PR #168, opened 2014-12-28
Fixed issue with double backslash preserving.

### LINKED ISSUE #167: Loader doesn't preserve double backslashes
It seems that loader doesn't preserve double backslashes.

```yaml
prop: |
    \\ these two backslashes become one in object
```

tyomo4ka:
Closes #167.


Seldaek:
Thanks

PR #180, opened 2015-02-19
Update README.md

ireinhart:
clarify usage of references with includes

### DIFF
--- a/README.md
+++ b/README.md
@@ -542,6 +542,31 @@ Nelmio\Entity\User:
 All files are merged in one data set before generation, and the includer's content
 takes precedence over included files' fixtures in case of duplicate keys.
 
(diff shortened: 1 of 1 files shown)


Seldaek:
This seems like a bug we should fix though instead of documenting the workaroud :)

ireinhart:
Did you have open an issue or pull request? In that case we can kill this pull request.

Seldaek:
No I figured we can keep this open as a reminder but if you rather open a clear bug report instead it would be even better :)

theofidry:
I can't reproduce it locally and I don't have it either in #340 (which add a few tests regarding the parsers). Try to update your version and if you still get this error re-open an issue for it.

Cheers

PR #188, opened 2015-03-17
Added new documentation for 2.0

tshelburne:
@Seldaek Updated changelog, upgrade, and readme (briefly) to describe 2.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+### 2.0.0 (2015-03-17)
+
+  * Expanded public interface of the `Loader` class to support:
+    - Trivial parsing of new document types
+    - Custom instantiation methods
+    - Custom setting of properties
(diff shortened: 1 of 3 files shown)


Seldaek:
@tshelburne please run with this and tag a 2.0 already :) It's quite obvious I don't have time to manage alice. I hoped I'd find time but it's not happening so go with what you feel is right, if it's wrong there's always a next release where things can be fixed!

Seldaek:
Also, as soon as 2.0 is out more people will start migrating which will reduce the 1.x PR/issue load, so that'd be good too.

tshelburne:
@Seldaek Ok - last we talked (admittedly a while ago), you wanted to see this before we went for it. Will take care of it.

Seldaek:
Yeah I wanted to but I have to let go of some things because I can't do everything :)

PR #189, opened 2015-03-21
fixed wrong loader instantiation for section "Detail Usage"

stevleibelt:


Seldaek:
This is actually correct for the upcoming 2.0 / master branch sorry for the confusion but you should check the 1.x readme if you use that :)

stevleibelt:
as long as it is solved :-)

Thanks for the work so far :+1:

PR #200, opened 2015-04-21
Added mention of FOSUserBundle listener enabled by default.

### LINKED ISSUE #199: Processor with FOSUserBundle
Could I have information on what the [example of processor](https://github.com/nelmio/alice#processors) given really does? Is it really necessary because I forgot to use it and it seemed to work so just asking to clear things up.

theofidry:
Closes #199.

### DIFF
--- a/README.md
+++ b/README.md
@@ -784,6 +784,8 @@ $loader->addProcessor($processor);
 $objects = $loader->loadFiles(__DIR__.'/fixtures.yml');
 ```
 
+Note: by default, FOSUserBundle has a property `fos_user.use_listener` set to true which already does this job. So do not use a processor for that unless you have disabled this listener. For more information on it, check the [offical doc](https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Resources/doc/user_manager.md).
+
 ## License ##
 
 Released under the MIT License, see LICENSE.


tshelburne:
@theofidry @Seldaek Maybe it's just me, but I don't feel like we need to document another package's functionality - the example given in the README isn't supposed to say "here's how to use the `FOSUserBundle`", it's to say "here's where you can drop code to handle processing of your objects". If it's that confusing, we could just change the code in the methods to be something more benign, like `$object->password = PasswordHasher::hash($object->password);`. Thoughts?

jenkoian:
@tshelburne yeah I think that's fair, my intention when I added the example was exactly that, an example. I only used fos user as an example as I thought it could prove a common scenario. If it's proving to hinder rather than help though, then feel free to update to something more generic :+1:

theofidry:
Either way I'm fine: either use another example or add a mention for the example because otherwise is misleading.

tshelburne:
@theofidry Fair point - are you going to update the PR, or do you want me to take care of it?

theofidry:
@tshelburne if it doesn't bother you I would prefer that you do it I have trouble to keep up with everything right now^^

tshelburne:
Alright, closing and updating independent of this.

theofidry:
Thanks @tshelburne.

PR #216, opened 2015-06-11
Allow arrays parameters

ogizanagi:
This allow to have array parameters. For instance:

```php
#fixtures.yml
parameters:
    usernames: ['Alice', 'Bob', 'Ryan']

Nelmio\Entity\User:
    user{1..50}:
        username: <randomElement(<{usernames}>)>
````

and associative arrays:
```php
parameters: 
    config: { host: localhost, port: 4567 }
```

Without this change, you'll get an `Array to string conversion` error:

However, note that it is already doable by enforcing the yaml array to be interpreted as a string:
```php
parameters:
    usernames: "['Alice', 'Bob', 'Ryan']"
```

### DIFF
--- a/src/Nelmio/Alice/Instances/Processor/Methods/Parameterized.php
+++ b/src/Nelmio/Alice/Instances/Processor/Methods/Parameterized.php
@@ -50,7 +50,11 @@ class Parameterized implements MethodInterface
                 ));
             }
 
-            return $this->parameters->get($key);
+            if (is_array($value = $this->parameters->get($key))) {
+                return var_export($value, true);
+            }
+
+            return $value;
         }, $value);
     }
 }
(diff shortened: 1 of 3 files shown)


Seldaek:
Could you add a test for it?

ogizanagi:
Tests added. One or the other should be sufficient, though.
Let me now if I should remove one.

Seldaek:
Thanks looks good to me, hopefully I am not missing anything

tshelburne:
:+1:

Perfect-Web:
does the associative array work ? i couldnt make it work no matter what

theofidry:
@Perfect-Web it works fine on 3.x I didn't check on 2.x. If it doesn't and you wish it to work, feel free to open a new issue for it or a PR, but don't dig zombies please :)

Perfect-Web:
@theofidry thats fine :) but still is on edge php 7 only sooo unless we change all the env, there is nothing to do ...

PR #221, opened 2015-06-19
Add KNPLabs/rad-fixtures-load to Symfony2 integrations

PedroTroller:
Hello guys.
Just a little add to the documentation to promote my [Symfony2 integration](https://github.com/KnpLabs/rad-fixtures-load) of your awesome library.
It's a configuration-less integration, let's try !

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ This is installable via [Composer](https://getcomposer.org/) as [nelmio/alice](h
 
     composer require nelmio/alice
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Please don't do things like requiring `"nelmio/alice": ">=1.5.0",` that's a very bad idea if we some day release a 3.0 breaking everything.

2.0 is fairly compatible, so if your bundle works with 1.5+ and 2.0 as well please use `1.5 - 2.99` or `^1.5 || ^2.0` as requirement, both of which will allow 2.* versions and 1.5+ in the 1.* range

Seldaek:
As for this PR, thanks ;)

PedroTroller:
@Seldaek Of course, https://github.com/KnpLabs/rad-fixtures-load/pull/7
Thanks.

PR #223, opened 2015-06-26
Fixed error with empty input

enumag:
With empty input I got this error:

```
Warning: array_merge() expects at least 1 parameter, 0 given
```


tshelburne:
Fair - out of curiosity, why are you passing empty input to this?

Please add a regression test, happy to merge afterwards.

enumag:
I'm new to Alice so I just ran into this randomly when my fixtures.yml file was empty.

I'll add the test soon.

enumag:
The test with custon CSV parser was failing for me so I fixed that as well (in separate commit). Please check if the fix is to your liking.

tshelburne:
Thanks, this is great!

PR #229, opened 2015-07-11
Add provider when no generators

theofidry:
As of now if we retrieve a fresh instance of `Nelmio\Alice\Instances\Processor\Methods\Faker` and call `::addProvider()`, [this foreach block](https://github.com/nelmio/alice/blob/master/src/Nelmio/Alice/Instances/Processor/Methods/Faker.php#L97-L99) will throw an error as `$this->generators` is still `null`.


tshelburne:
Please add a regression test and we're good to go.

PR #232, opened 2015-07-12
Moved to PSR-4

theofidry:
* flattened `src` and `tests` (removed `Nelmio/Alice`)
* used `autoload-dev` in `composer.json` to avoid to declare a custom `tests/bootstrap.php`


Seldaek:
+1 for the autoload-dev change.

The rest I don't think it brings anything except breaking open PRs and potentially creating conflicts when trying to merge up from the 1.0 branch so -1 for me.

theofidry:
Regarding the PR out of the 3 other opens, 2 of them are mine, but if you prefer to keep it I'll just put the autoload-dev change :)

Seldaek:
Yeah it's not a huge deal I know but I also don't really see the advantage of skipping those two dirs.

theofidry:
urg, does not seem to really work, I guess there is a kind of conflict since the two autoload statements are using the same prefix.

theofidry:
Done

Seldaek:
What was the issue? Because conflicting prefixes should not be one it should just try both. Anyway looks fine now.

theofidry:
it was just me in local, I didn't reupdated the dependencies so the autoloader wasn't updated :)

Seldaek:
Ok, `composer dump` ;)

PR #234, opened 2015-07-22
Fixing error when the parameter is got by func_get_args

rafreis:
When I have:

```yml
MyNameSpace\Object (local):
    obj1:
        __construct: ['val']
```

And My Object is:

```php
<?php

namespace MyNameSpace;

class Object
{
    public function __construct()
    {
        $args = func_get_args();
        // do something
    }
}
```

I got error: Notice: Undefined offset: 0 (Symfony\Component\Debug\Exception\ContextErrorException)

With this patch this behavior will be fixed.

### DIFF
--- a/src/Nelmio/Alice/Util/TypeHintChecker.php
+++ b/src/Nelmio/Alice/Util/TypeHintChecker.php
@@ -50,7 +50,7 @@ class TypeHintChecker
         $reflection = new \ReflectionMethod($obj, $method);
         $params = $reflection->getParameters();
 
-        if (!$params[$pNum]->getClass()) {
+        if (!isset($params[$pNum]) || !$params[$pNum]->getClass()) {
             return $value;
         }
 


jefersondaniel:
+1

jpizzribeiro:
+1

tshelburne:
@rafreis Sorry for the wait - this is great, but please add a regression test.

rafreis:
@tshelburne I believe that Is not necessary 'cause every unity tests already is passing and, is just a check if the constructor method accept parameters or not.

tshelburne:
@rafreis I'm not worried about it causing existing tests to fail - if I merged your PR, and two months down the line deleted that line, all the tests would *still* pass. The test showing the problem you are currently having is necessary to ensure we don't remove something and break your code down the road.

tshelburne:
Closed by #238 - thanks for catching this!

PR #238, opened 2015-08-17
Fixing error when the parameter is got by func_get_args

jefersondaniel:
Testing for PR #234


rafreis:
@jefersondaniel tks ;)

Issue #242, opened 2015-08-18
Inject Fixtures loaders


theofidry:
As of now, when calling `Nelmio\Alice\Fixtures::loadFiles()`, the loader is resolved with the following function:

```php
private static function getLoader(array $options)
    {
        // Generate an array key based on the options, so that separate loaders
        // will be created when we want to load several fixtures that use different
        // custom providers.
        $loaderKey = self::generateLoaderKey($options);
        if (!isset(self::$loaders[$loaderKey])) {
            self::$loaders[$loaderKey] = new Loader($options['locale'], $options['providers'], $options['seed']);
        }

        return self::$loaders[$loaderKey];
    }
```

But what if we want to use a custom Loader? A brand new loader or for example a `Nelmio\Alice\Fixtures\Loader` which has been initialised with some parameters.

tshelburne:
`Fixtures`, as of right now, is basically just the path of least resistance. It's not intended to be the flexible answer, it just works for the simplest possible situation. I personally never use it.

From my perspective, the `Fixtures` is little more than a wrapper around `Loader`, with some conveniences thrown into the constructor. As soon as any custom behavior is needed, I suggest just using `Loader` or writing your own extension of `Loader`. 

Thoughts?

theofidry:
So to recap, `Loader` does the actual loading to return the objects to persist. `Fixtures` is a bootstrap to take care of the persistence. I took a deeper look at `Fixtures` and from what I see its complexity (I don't mean it's a very complex object) comes from:
* the fact that the loader is not injected by instantiated in it
* the actual loading in which the loader is called and the objects retrieved are persisted.

The issue is that the loader is instantiated and there's not much that can be done since we're in a static context, unless we pass the loader in the `::load()` function signature.

Back to our cases: having a custom `Fixtures` to handle the loader makes sense, what I'm not happy with though is that I'm using the exact same function `::load()` so it's a plain copy/paste.

So here's my proposal:
1. We let things as they are and we just add a chapter in the doc regarding this use case
2. We add another `Fixtures` extension, which keeps the persistence logic as in the current `Fixtures`, but which does not aim to be used in a static context allowing to inject our `Loader`, custom or not, in it. This new extension would not be actually used in the library, but allow people to use their own loaders without having to redeclare a new extension.

I would like 2. of course, it opens up the library without being much to maintain. But obviously, if you can prefer to let this to the initiative of the users and choose 1. If you're fine with 2, I can take care of the PR.

[...]

PR #243, opened 2015-08-18
Added FixturesLoader

theofidry:
@tshelburne follow up of #242.

### DIFF
--- a/src/Nelmio/Alice/FixturesLoader.php
+++ b/src/Nelmio/Alice/FixturesLoader.php
@@ -1,0 +1,132 @@
+<?php
+
+/*
+ * This file is part of the Alice package.
+ *
+ * (c) Nelmio <hello@nelm.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\Alice;
+
+use Doctrine\Common\Persistence\ObjectManager;
+use Nelmio\Alice\Fixtures\Loader;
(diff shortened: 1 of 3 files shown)


theofidry:
@tshelburne @Seldaek any though on this?

sgb-io:
+1

theofidry:
@Seldaek if you could check this too, would be great :)

src/Nelmio/Alice/FixturesLoader.php
```diff
@@ -0,0 +1,132 @@
...
+     * @param string|array $files filename, glob mask (e.g. *.yml) or array of filenames to load data from, or data
+     *                            array
+     * @param bool         $persist_once
+     *
+     * @return array Persisted objects.
+     */
+    public function load($files, $persist_once = false)
+    {
+        // glob strings to filenames
+        if (!is_array($files)) {
```
theofidry: I don't really like this part: L79-L86. It's about checking if the fixtures files really exists, which is IMO not the responsibility of the loader.

Seldaek:
I'd rather have @tshelburne look over this one because I am not so familiar with the 2.0 bootstrapping code tbh, haven't used alice myself in a while.

tshelburne:
@theofidry Happy to check out all these updates soon - I've been in Iceland for 2 weeks now, heading home in a few days. Will make a point of digging in then.

theofidry:
Cool man, enjoy your holidays :+1: for now, got much to discuss after.

tshelburne:
@theofidry See comment in #242

Issue #244, opened 2015-08-18
Inconsistant interface for PersisterInterface::find() when no result found


theofidry:
The [`Nelmio\Alice\Persister\PersisterInterface::load()`](https://github.com/nelmio/alice/blob/master/src/Nelmio/Alice/PersisterInterface.php#L30) is very unclear about the result return. It is used only in two places:
* for the doctrine bridge: [`Nelmio\Alice\Persister\Doctrine::load()`](https://github.com/nelmio/alice/blob/master/src/Nelmio/Alice/Persister/Doctrine.php#L48) which throws an `\UnexpectedValueException` exception when no result is found
* [`Nelmio\Alice\TestPersister`](https://github.com/nelmio/alice/blob/master/tests/Nelmio/Alice/TestPersister.php#L33) which suggest it returns null when no result is found

As a Doctrine user, I feel more right to return `null` when no  result is found but throwing an error makes sense too. But it would be better to make a choice on this.

tshelburne:
I agree with null.

theofidry:
Ok I'll do a PR to reflect on that then :)

theofidry:
Follow up in #245.

PR #245, opened 2015-08-20
Rework the Persister interface and bridge

theofidry:
@tshelburne follow up of #244.

### DIFF
--- a/src/Nelmio/Alice/Persister/Doctrine.php
+++ b/src/Nelmio/Alice/Persister/Doctrine.php
@@ -51,7 +51,7 @@ class Doctrine implements PersisterInterface
             $this->manager->persist($object);
         }
 
-        if (true === $this->flush) {
+        if ($this->flush) {
             $this->manager->flush();
         }
     }
--- a/src/Nelmio/Alice/Persister/Doctrine.php
+++ b/src/Nelmio/Alice/Persister/Doctrine.php
@@ -15,17 +15,31 @@ use Doctrine\Common\Persistence\ObjectManager;
 use Nelmio\Alice\PersisterInterface;
 
 /**
- * The Doctrine persists the fixtures into an ObjectManager
+ * Adaptor for the ObjectManager to provide a Doctrine bridge.
+ *
+ * @author Jordi Boggiano <j.boggiano@seld.be>
+ * @author Robert Schönthal
  */
 class Doctrine implements PersisterInterface
 {
-    protected $om;
+    /**
+     * @var ObjectManager
+     */
+    protected $manager;
+
+    /**
+     * @var bool
+     */
     protected $flush;
 
-    public function __construct(ObjectManager $om, $doFlush = true)
+    /**
+     * @param ObjectManager $manager
+     * @param bool          $flush
+     */
+    public function __construct(ObjectManager $manager, $flush = true)
     {
-        $this->om = $om;
-        $this->flush = $doFlush;
+        $this->manager = $manager;
+        $this->flush = $flush;
     }
 
     /**
@@ -34,11 +48,11 @@ class Doctrine implements PersisterInterface
     public function persist(array $objects)
     {
         foreach ($objects as $object) {
-            $this->om->persist($object);
+            $this->manager->persist($object);
         }
 
-        if ($this->flush) {
-            $this->om->flush();
+        if (true === $this->flush) {
+            $this->manager->flush();
         }
     }
 
@@ -47,12 +61,6 @@ class Doctrine implements PersisterInterface
      */
     public function find($class, $id)
     {
-        $entity = $this->om->find($class, $id);
-
-        if (!$entity) {
-            throw new \UnexpectedValueException('Entity with Id ' . $id . ' and Class ' . $class . ' not found');
-        }
-
-        return $entity;
+        return $this->manager->find($class, $id);
     }
 }
--- a/src/Nelmio/Alice/PersisterInterface.php
+++ b/src/Nelmio/Alice/PersisterInterface.php
@@ -11,21 +11,28 @@
 
 namespace Nelmio\Alice;
 
+/**
+ * The persister is the class responsible for persisting objects into the database.
+ *
+ * @author Jordi Boggiano <j.boggiano@seld.be>
+ * @author Robert Schönthal
+ */
 interface PersisterInterface
 {
     /**
-     * Loads a fixture file
+     * Persists objects to database.
      *
-     * @param array[object] $objects instance to persist in the DB
+     * @param object[] $objects
      */
     public function persist(array $objects);
 
     /**
-     * Finds an object by class and id
+     * Finds an object by class and id.
      *
      * @param  string $class
      * @param  int    $id
-     * @return mixed
+     *
+     * @return object|null null if object not found.
      */
     public function find($class, $id);
 }


src/Nelmio/Alice/Persister/Doctrine.php
```diff
@@ -15,17 +15,31 @@
...
+     * @var bool
+     */
     protected $flush;
 
-    public function __construct(ObjectManager $om, $doFlush = true)
+    /**
+     * @param ObjectManager $manager
+     * @param bool          $flush
+     */
+    public function __construct(ObjectManager $manager, $flush = true)
```
theofidry: Is the `flush` parameter necessary? It should always be `true` no?

theofidry:
Also I find the namespacing a little bit weird, could we change it like this?
* `Nelmio\Alice\PersisterInterface\Doctrine` => `Nelmio\Alice\Doctrine\Persister\Persister`
* `Nelmio\Alice\PersisterInterface` => `Nelmio\Alice\Persister\PersisterInterface`

To place persisters in the `Nelmio\Alice\Doctrine\Persister` namespace and separate the Doctrine bridge into a dedicated namespace.

To avoid BC break, it could be done by "duplicating" it and adding some deprecation notice and remove the deprecated classes for the `3.x`.

src/Nelmio/Alice/Persister/Doctrine.php
tshelburne: I don't have an opinion on this - I personally haven't used the built-in persistence part of Alice. @Seldaek ?

src/Nelmio/Alice/Persister/Doctrine.php
```diff
@@ -47,12 +61,6 @@ public function persist(array $objects)
      */
     public function find($class, $id)
     {
-        $entity = $this->om->find($class, $id);
-
-        if (!$entity) {
```
tshelburne: I'm assuming this is because you have added `object|null` in the interface doc?

src/Nelmio/Alice/Persister/Doctrine.php
```diff
@@ -34,11 +48,11 @@ public function __construct(ObjectManager $om, $doFlush = true)
     public function persist(array $objects)
     {
         foreach ($objects as $object) {
-            $this->om->persist($object);
+            $this->manager->persist($object);
         }
 
-        if ($this->flush) {
-            $this->om->flush();
+        if (true === $this->flush) {
```
tshelburne: Not sure that this is really necessary, since `$this->flush` isn't used in a way where truthy / falsy values would cause problems.

src/Nelmio/Alice/Persister/Doctrine.php
theofidry: What I meant there was that rather changing the behaviour of the persist, we should keep the flush separately. I haven't checked if it was used anywhere but seems not

src/Nelmio/Alice/Persister/Doctrine.php
theofidry: Yep. I don't personally have a string opinion about wether an exception should be thrown or null returned, but it's better to stay consistent

tshelburne:
I am in favor of moving towards a "Doctrine bridge" like solution in the short term, since I think it would be best for us to move towards pulling that out of Alice anyway. I've got nothing against your proposal, but would it make sense to make that a separate issue under a 3.0 milestone, rather than bundling it into a PR and having duplicated code?

src/Nelmio/Alice/Persister/Doctrine.php
theofidry: ? You mean the type check or the flush thing?

src/Nelmio/Alice/PersisterInterface.php
```diff
@@ -11,21 +11,28 @@
...
+ * @author Robert Schönthal
+ */
 interface PersisterInterface
 {
     /**
-     * Loads a fixture file
+     * Persists objects to database.
      *
-     * @param array[object] $objects instance to persist in the DB
+     * @param object[] $objects
```
theofidry: Err, does the persist have to return something actually

theofidry:
> rather than bundling it into a PR and having duplicated code?

You're talking of that right?

> Also I find the namespacing a little bit weird, could we change it like this?
>
>Nelmio\Alice\PersisterInterface\Doctrine => Nelmio\Alice\Doctrine\Persister\Persister
Nelmio\Alice\PersisterInterface => Nelmio\Alice\Persister\PersisterInterface
To place persisters in the Nelmio\Alice\Doctrine\Persister namespace and separate the Doctrine bridge into a dedicated namespace.
>
>To avoid BC break, it could be done by "duplicating" it and adding some deprecation notice and remove the deprecated classes for the 3.x.

It would add a little bit of duplication but not much actually: adding an interface, extending the other and adding a deprecation notice. It adds a little bit of useless code but would give the opportunity to the users to see the change coming.

Another benefit is that if we choose to put that in a 3.x only, then it requires to create a new branch now and be wary of synchronising the two branches. That would be fine if the 2.x was pretty stable of the 3.x well advanced. But for something so trivial I'm not sure it's really worth it.

That being said, that depends of how much energy you want to put into it :p So your call man.

> I am in favor of moving towards a "Doctrine bridge" like solution in the short term, since I think it would be best for us to move towards pulling that out of Alice anyway.

I'm not sure of what you mean. I like the `Persister` idea as it loosen up the coupling. Do you have something else in mind?

src/Nelmio/Alice/Persister/Doctrine.php
Seldaek: I think strict checking for true here is harmful indeed as it might break things that passed in 1 or whatever truthy/falsey value.

src/Nelmio/Alice/Persister/Doctrine.php
Seldaek: I think in v1 the point of this was that you could disable the automatic flush if you wanted to call persist/load many times and then flush them all at once. It's never been used in alice itself but was an option for users to use. Whether it still makes sense in the new code I am not sure.

src/Nelmio/Alice/Persister/Doctrine.php
tshelburne: I'm fine with returning null in general, but this *is* a breaking change. Anyone catching the exception would have their code break, so I don't think we can just do it "lightly".

src/Nelmio/Alice/Persister/Doctrine.php
tshelburne: This makes sense as a feature to leave in from my perspective, but again, I don't use it.

src/Nelmio/Alice/Persister/Doctrine.php
theofidry: Erf true. Should we let the exception for now and add a deprecation notice?

src/Nelmio/Alice/Persister/Doctrine.php
theofidry: As a Doctrine user I find that more confusing than anything. But then again changing this causes some BC break. Not in the library itself neither for people using `Fixtures`. So theatrically it may break something for someone.

We can also keep the current behaviour for now and add a deprecation notice here too (I'm getting the feeling I'm only talking about deprecation tonight).

tshelburne:
Upon thinking more about it, I think it would be better to just leave persistence isolated in a single directory for now. If we see Doctrine having more than a single file (the persister) down the road, perhaps we can make a subdirectory under the Persister folder.

Jordi and I have discussed pulling persistence out of Alice and instead creating some "bridge" libraries that could be installed and used independently. Not hugely important, but it would limit the drive to "put more" into Alice. It would mean defining the intention of Alice a bit more strictly, in that it would exclusively be an object generation library, but I think it would bring a lot of clarity to it as well.

tshelburne:
My feeling is that this PR should probably be closed as it is for now, since there are so many BC issues, and we can continue discussion in relevant issues. Make sense?

theofidry:
> Jordi and I have discussed pulling persistence out of Alice and instead creating some "bridge" libraries that could be installed and used independently. Not hugely important, but it would limit the drive to "put more" into Alice. It would mean defining the intention of Alice a bit more strictly, in that it would exclusively be an object generation library, but I think it would bring a lot of clarity to it as well.

That would be a good move, and if you wish to advance to this one I'll gladly help but then create a 2.x branch :p

> My feeling is that this PR should probably be closed as it is for now, since there are so many BC issues, and we can continue discussion in relevant issues. Make sense?

Let it, I'll remove the BC and just let the fix in the phpdoc.

tshelburne:
If you want to do it this way, that's fine, but let's squash the commits down to a single one - I don't see a benefit to having a commit in the upstream repo with breaking changes. If it's about history, I think it would be better documented in issues.

theofidry:
Yeah of course, but let me fix the BC things first ;)

theofidry:
Closing in favour of #245

PR #247, opened 2015-08-22
Update README.md

ma-si:
Add third-party library link to Zend Framework 2 module


theofidry:
@ma-si needs a little rebase ;)

PR #248, opened 2015-08-25
Minor typo fix

theofidry:
* use `$manager` instead of `$om`
* use short array notation (available since PHP 5.4 and some other example in the doc was with the short array notation)


theofidry:
@tshelburne is back :dancer:

PR #252, opened 2015-09-05
Fix container exception

theofidry:
When using another persister than `Doctrine\Common\Persistence\ObjectManager`, we get the exception bits `\InvalidArgumentException('Unknown container type '.get_class($this->container))`. It's really weird to have that when we have a `PersisterInterface` especially if under the hook there is not other support than the `ObjectManager`.

So instead, we can require the `persister` to implement the `PersisterInterface` and for not introducing any BC break (in the interface) keep accepting the `ObjectManager` in the `::load()` method. This way, if we happen to have another persister, we can implement our own bridge.

On the way I renamed `container` to `persister` as is was already a bit unclear and now would be inconsistent.

### DIFF
--- a/src/Nelmio/Alice/Fixtures.php
+++ b/src/Nelmio/Alice/Fixtures.php
@@ -12,62 +12,108 @@
 namespace Nelmio\Alice;
 
 use Doctrine\Common\Persistence\ObjectManager;
+use Nelmio\Alice\Persister\Doctrine;
 use Psr\Log\LoggerInterface;
 use Nelmio\Alice\Fixtures\Loader;
-use Nelmio\Alice\Persister\Doctrine as DoctrinePersister;
 
 class Fixtures
 {
+    /**
+     * @var Loader[]
+     */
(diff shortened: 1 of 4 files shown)


Seldaek:
Seems fine to me but it's breaking the test suite, can you check that?

theofidry:
Hm indeed, plus I didn't add much tests on it, I'm fixing that

theofidry:
This is due to the use of the `fixtures` key in the options in a test suite: any idea of its use?

Seldaek:
Err, it was to provide a list of fixture files before I think, not sure if this was removed in 2.0 or what exactly happened.

theofidry:
Fixed. As the `fixtures` key was never checked and used I removed it.

Seldaek:
Thanks

Seldaek:
Ah I see it was used to have a fixture as a php array instead of loading it from a yml file..

theofidry:
Seems like it. But no longer used

PR #255, opened 2015-09-07
Fix logger setting

meyerbaptiste:
Since https://github.com/nelmio/alice/commit/a2cb638c1a29d9335a46b779c5d77dbd9d7f00a6 the logger of the loader is not set from the `Fixtures` class.


Seldaek:
Thanks

PR #258, opened 2015-09-11
Use preg_grep to match collection instances by mask

munkie:
*preg_grep* is much more faster alternative than iterating collection and applying *preg_match* on every instance array key.
I've used this script to random method performance  - https://gist.github.com/munkie/1d06214d8eaa2af349c1
Performance boost was upto 3-4x times with 1000 and 10000 items counts.

### DIFF
--- a/src/Nelmio/Alice/Instances/Collection.php
+++ b/src/Nelmio/Alice/Instances/Collection.php
@@ -134,12 +134,12 @@ class Collection
             return [];
         }
 
-        $availableObjects = [];
-        foreach ($this->instances as $name => $instance) {
-            if (preg_match('{^'.str_replace('*', '.+', $mask).'$}', $name)) {
-                $availableObjects[] = $name;
-            }
-        }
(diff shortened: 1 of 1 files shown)


tshelburne:
@munkie Thanks for this! Can you separate these additions out into individual PRs? My suggestion would be the `preg_grep` change as one PR, the fix in `Reference` with the test update in another, and finally the changes in `LoaderTest` in a third - this will allow us to discuss them independently, since they really aren't related.

munkie:
@tshelburne I've splitted this PR into three:
* This PR now only introduces usage of _preg_grep_ to improve _Collection::random_ performance
* #260 LoaderTest cleanup
* #261 Add test for 0x multiplier in reference

theofidry:
Cool, thanks @munkie!

munkie:
@tshelburne I can provide same fix for 1.x branch, is it needed ?

tshelburne:
Thanks!

PR #260, opened 2015-09-14
Cleanup LoaderTest test

munkie:
Clean up LoaderTest
* Remove unused persister
* Remove unused local vars
* Use _setExpectedExcetion_ method for exception assertions

Also i would recommend to:
* Remove composer.lock it has outdated versions of dependencies.
* Use last stable version of phpunit (~4.8)

### DIFF
--- a/tests/Nelmio/Alice/Fixtures/LoaderTest.php
+++ b/tests/Nelmio/Alice/Fixtures/LoaderTest.php
@@ -11,9 +11,6 @@
 
 namespace Nelmio\Alice\Fixtures;
 
-use Nelmio\Alice\support\models\PrivateConstructorClass;
-use Nelmio\Alice\support\models\Group;
-use Nelmio\Alice\TestPersister;
 use Nelmio\Alice\support\models\User;
 use Nelmio\Alice\support\extensions;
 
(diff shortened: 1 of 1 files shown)


tests/Nelmio/Alice/Fixtures/LoaderTest.php
```diff
@@ -94,11 +88,12 @@ public function testGetBadReference()
 
     public function testLoadUnparsableFile()
     {
-        try {
-            $res = $this->createLoader()->load($file = __DIR__.'/../support/fixtures/not-parsable');
-        } catch (\UnexpectedValueException $e) {
-            $this->assertEquals("{$file} cannot be parsed - no parser exists that can handle it.", $e->getMessage());
-        }
+        $file = __DIR__.'/../support/fixtures/not-parsable';
+        $this->setExpectedException(
```
tshelburne: Does our version of PHPUnit not have the support for annotated exceptions?

tshelburne: Actually, I see that we're using annotations elsewhere - is there any reason you're choosing not to here?

munkie: Because $file needs to be added to the beginning of expected exception message

tshelburne: Oh... wow, yea, good call. Attention to detail...

PR #262, opened 2015-09-14
Use preg_grep to filter references by mask

munkie:
Fixes same random references performance issue as #258 but for 1.x branch


tshelburne:
Thanks!

Issue #241, opened 2015-08-17
Add Scrutinizer and SensioLabs Insight?


theofidry:
Since they're both free for public open source projects, it would be great to have them!

theofidry:
Waiting for https://github.com/scrutinizer-ci/scrutinizer/issues/382

PR #263, opened 2015-09-14
Add ProcessableInterface::getMatches

munkie:
_Faker::process_ method expects $processable to be _ProcessableInterface_, but uses not-existing property _matches_. To solve this issue i've added getMatches method to ProcessableInterface interface.

### DIFF
--- a/src/Nelmio/Alice/Instances/Processor/ProcessableInterface.php
+++ b/src/Nelmio/Alice/Instances/Processor/ProcessableInterface.php
@@ -33,4 +33,11 @@ interface ProcessableInterface
      * @return string
      */
     public function getMatch($name);
+
+    /**
+     * return all matches
+     *
+     * @return array|string[]
+     */
+    public function getMatches();
 }
--- a/src/Nelmio/Alice/Instances/Processor/ProcessableInterface.php
+++ b/src/Nelmio/Alice/Instances/Processor/ProcessableInterface.php
@@ -37,7 +37,7 @@ interface ProcessableInterface
     /**
      * return all matches
      *
-     * @return array|string[]
+     * @return string[]
      */
     public function getMatches();
 }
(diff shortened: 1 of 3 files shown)


src/Nelmio/Alice/Instances/Processor/ProcessableInterface.php
```diff
@@ -33,4 +33,11 @@ public function valueMatches($regexString);
      * @return string
      */
     public function getMatch($name);
+
+    /**
+     * return all matches
+     *
+     * @return array|string[]
```
theofidry: why both `array` and `string[]`?

munkie: OK, i can leave just _string[]_

theofidry: Well, actually I don't know, if its jut `string[]` yeah, if it can return something else we have to specify it too, but `array` is way too unclear

tshelburne:
Can you confirm that `Faker::process` is the only place where matches was used as a property? I like this fix, but I want it to cover everything.

src/Nelmio/Alice/Instances/Processor/ProcessableInterface.php
munkie: It return matches created by preg_match and they are always strings.

munkie:
@tshelburne I confirm, have not found any other PropertyInterface::$matches usage.
By the way, what do you think about using https://scrutinizer-ci.com/ for code analysis ?

theofidry:
> By the way, what do you think about using https://scrutinizer-ci.com/ for code analysis ?

:+1: #241

tshelburne:
I haven't used it before - when I joined the project as a maintainer, Jordi already had the CS stuff set up. I'm not against it by any means, but that would definitely take his input before merging.

munkie:
@tshelburne I've created an account for my fork so you can check what kind of reports it has - https://scrutinizer-ci.com/g/munkie/alice/

tshelburne:
Yea, that's pretty awesome - thanks for the suggestion. Spike something out.

PR #266, opened 2015-09-14
Fix actual return type of methods where preg_match is used to be bool not int

munkie:


src/Nelmio/Alice/Fixtures/Parser/Methods/Base.php
```diff
@@ -39,7 +39,7 @@ public function canParse($file)
     {
         // we add (\.php)? to the regex to allow extensions of this parser to first
         // be compiled by php
-        return preg_match("/\.{$this->extension}(\.php)?$/", $file) == 1;
+        return 1 === preg_match("/\\.{$this->extension}(\\.php)?$/", $file);
```
tshelburne: Why the double "\"?

munkie: Backslash is used to escape special characters, thou there is no \. sequence, it's safier to use doubled backslash
> To specify a literal backslash, double it (\\)

See double quotes string type docs - http://php.net/manual/en/language.types.string.php#language.types.string.syntax.double

tshelburne: Ah, missed the double quotes. Ok, looks good.

PR #268, opened 2015-09-14
Collection improvements

munkie:


theofidry:
So, `sprintf()` over strings concatenation. I personally prefer but I'm curious to know if you have any argument for it

munkie:
Well, it is more readable than concatenated one. Also symfony recommends to use sprintf for exception messages.
> Exception message strings should be concatenated using sprintf.
See - http://symfony.com/doc/current/contributing/code/standards.html#structure

theofidry:
true, but its not a Symfony project ;)

tshelburne:
I don't have an opinion on this - @Seldaek, do you mind either way?

Seldaek:
Nope i dont really mind. I think both have cases where they are more or
less readable and would advise common sense over blindly following rules,
as always.

Seldaek:
For what it's worth after looking at the diff I don't think these are great
sprintf cases. If you have only one param a string concat is shorter and
clear enough IMO. I'm not gonna fight it either way because it's a detail,
but change for change isn't really valuable.

theofidry:
There is just the empty check that I find cleaner ;)

munkie:
@Seldaek So should i revert sprintf's and just leave _empty_ check ?

theofidry:
@munkie for what it's worth you may let it... whatever we're not gonna fight over that :dancer:

PR #269, opened 2015-09-16
Cache keys by mask in instance collection

munkie:
This PR introduces one more performance optimization in Collection::random method

For example for following fixture:
```yaml
User:
  user{1..100}:
    name: <name()>
Group:
  group{1..100}:
    users: 5x user*
```
Collection::random method will be called 100 times and will create list of keys matching 'user*' mask 100 time and every time it will create same list of key.

I've added caching of keys matching mask, so keys list for mask will calculated only one time instead of 100.

### DIFF
--- a/src/Nelmio/Alice/Instances/Collection.php
+++ b/src/Nelmio/Alice/Instances/Collection.php
@@ -25,6 +25,17 @@ class Collection
     private $instances;
 
     /**
+     * Instance keys that match random mask
+     * [
+     *      mask => [keys matching mask],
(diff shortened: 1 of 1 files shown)


tshelburne:
Did you do any performance check on this? Curious how much performance improvement this actually gives.

munkie:
About 2x boost in this test https://gist.github.com/munkie/1d06214d8eaa2af349c1 for 1000 items

tshelburne:
Damn, that's awesome. I'm going to need to update my version of Alice shortly.

PR #272, opened 2015-09-19
Don't persist embeddable ORM entities

NiR-:


Einenlum:
Up ?

NiR-:
@Seldaek Only sf3 tests are broken here (cf. #290).

src/Nelmio/Alice/Persister/Doctrine.php
```diff
@@ -55,4 +60,21 @@ public function find($class, $id)
 
         return $entity;
     }
+
+    private function getPersistableClasses()
```
theofidry: it looks like this operation could be cached somehow and not being called at each persist.

NiR-: I'm not sure about what you mean here: persistable classes are already stored in a property after initialization.
Do you suggest to move the initialization in the constructor and use the property in `persist` without additional method call ?

theofidry: ah nvm, you're doing that in the function.

Einenlum:
Can we reload tests?

NiR-:
@Seldaek Can you replay travis build ? Tests against symfony3 should be fixed, thanks to your merge of #290.

Seldaek:
The failing test isn't the main problem.. we need someone to review :) Sorry but I don't have much time atm.

Einenlum:
Up?

theofidry:
@NiR- could you rebase it somehow to run the test with the latest version again?

NiR-:
@theofidry done.

theofidry:
@tshelburne ok for me

tshelburne:
@NiR- Thanks for this, sorry for wait - just finished a cross-country move this week.

matt-usurp:
Not meaning to Necro, but looking at the code this could work if the setter is dumb and doesn't have type-hinting, but if it doesn't have type-hinting then you will receive and error from the `find()` method against the `Doctrine` (`PersisterInterface`).

PR #282, opened 2015-10-05
Improve Property populator

enumag:
The Property populator currently can set values to private properties of the class itself but not to private properties of parent classes. It caused unexpected behaviour for me when I refactored one of my doctrine entitites to multiple entities with common parent.

Let me know if this is wanted in Alice. I'll add test in that case.


tshelburne:
My personal opinion (and unrequested advice): A private property is private because it's not supposed to be accessible to the outside world - when I'm making fixture files, I always try to use them to build objects the way I intend them to be built in the application itself. Setting private properties is asking for a mess if you depend on them being private for state validity.

That said, the fact that private properties *can* be set on instances, but only on the child class, is silly. Add the test and we'll merge it in.

enumag:
Thanks for your advice although I'm aware of it myself. I just don't know how else I can configure fixtures for my entities. If you can advice me about that I'd be greatful.

Here is the the problematic part of my entities:

```php
/**
 * @ORM\Entity
 */
class Demand
{
	// some other properties + their getters & setters

	/**
	 * @ORM\OneToOne( targetEntity = "DemandSchedule", inversedBy = "demand")
	 * @var DemandSchedule
	 */
	private $schedule;

	/**
	 * @param bool $need
	 * @return DemandSchedule
	 */
	public function getSchedule($need = false)
	{
		if ($need && !$this->schedule) {
			$this->schedule = new DemandSchedule($this);
		}
		return $this->schedule;
	}
}

/**
 * @ORM\Entity
 */
class DemandSchedule
{
	// some other properties + their getters & setters

	/**
	 * @ORM\OneToOne( targetEntity = "Demand", mappedBy = "schedule" )
	 * @var Demand
	 */
	private $demand;

	public function __construct(Demand $demand)
	{
		$this->demand = $demand;
	}

	/**
	 * @return Demand
	 */
	public function getDemand()
	{
		return $this->demand;
	}
}
```

I need a fixture for Demand (with schedule), currently I've this.

```
App\Entity\Demand:
    demand1:
        schedule: @schedule1 // this is the problem - setting private property
        // other fields

App\Entity\DemandSchedule:
    schedule1:
        __construct: false // to avoid error
        // other fields
```

Is there some other way to configure fixtures for my entities without relying on setting private property and obviously without adding `Demand::setSchedule($schedule)` method?

tshelburne:
Part of your struggle, I think, is in the fact that your getter is acting as a setter. You are attempting to make things work in ways that I personally would avoid. Your intuition in the last code snippet you pasted is correct - you really should be passing a demand to the demand schedules via the constructor. My suggestion, if it would be too much work to reorganize your data structure, would be to add a short reconciliation pass following creation of your entities. In other words, after you get your objects out, do something akin to the following (apologies for pseudo code, I've been in Purescript / Ruby world for months now):

```
$objects = $loader->load($fixtures);
foreach($objects as $object) {
  if ($object instanceof DemandSchedule) {
    // save relationship between $object and $object->getDemand();
  }
}
```

This may feel like a nasty solution, but it does solve your problem without requiring setting private properties, depending on a framework solution, or fixing a much larger problem (database design). In my projects that use Alice, I have many such reconciliation steps to deal with different ORMs, environmental requirements, and business rules.

enumag:
Test added. While working on it I also noticed that composer.lock is commited as well for some reason and that PHPUnit is way outdated so I fixed that as well.

tests/Nelmio/Alice/support/models/PropertiesParent.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Nelmio\Alice\support\models;
+
+class PropertiesParent
```
tshelburne: Let's leave this as `PluralProperties` - I don't really see a benefit to making this change, and it makes it look important if it's part of the history. Also, it's marking the other class as a modification of the existing class, instead of as a new class. Just generally confusing. Make sense?

tests/Nelmio/Alice/Instances/Populator/PopulatorTest.php
```diff
@@ -22,7 +23,7 @@
 class PopulatorTest extends \PHPUnit_Framework_TestCase
 {
     const CONTACT = 'Nelmio\Alice\support\models\Contact';
-    const PLURAL = 'Nelmio\Alice\support\models\PluralProperties';
+    const PROPERTIES = 'Nelmio\Alice\support\models\Properties';
```
tshelburne: I would rather explicitly have the child class used than implicitly use the child class for both test cases.

tests/Nelmio/Alice/support/models/PropertiesParent.php
enumag: Sure, it just didn't make sense to use a class called PluralProperties for test that doesn't work with arrays. But as you wish, I'll change it back.

tests/Nelmio/Alice/Instances/Populator/PopulatorTest.php
enumag: You mean to remove these constants?

tshelburne: No, sorry that was unclear - I would rather see constants for both the child class and the parent class and use them explicitly in the tests that require them so that it's clear what the intention is. For that matter, the `testProperty` method should probably be called `testSettingPrivatePropertiesDirectly`. Thoughts?

enumag: No test uses the parent class directly, in both cases the child class is needed.

I just used the existing naming scheme for the method, will rename as you wish.

enumag:
@tshelburne All done, can you review it again? :-)

PR #287, opened 2015-11-06
Allow this component to work with symfony3

benji07:


.travis.yml
```diff
@@ -7,6 +7,26 @@ php:
...
+    # Test against LTS releases
+    - php: 5.5
+      env: SYMFONY_VERSION='2.3.*'
+    - php: 5.5
+      env: SYMFONY_VERSION='2.7.*'
+    # Test against dev dependencies
+    - php: 5.6
+      env: DEPS=dev
+    - php: 5.6
+      env: DEPS=dev SYMFONY_VERSION='2.8.*'
```
Soullivaneuh: `2.8.*@dev` would be better than a `DEPS` option IMO.

tshelburne:
@Seldaek It's been a bit since I dealt with Symfony and PHP - does this all look good to you?

Seldaek:
Seems fine yeah, we barely use any symfony anyway. One issue though is that AFAIK symfony3's yaml component doesn't allow unquoted `@`s anymore so all our `@foo` references will need to become `'@foo'`, as the @ is a reserved char in the yaml spec. This might not break the tests, but I think we should update docs accordingly, and well users migrating to sf3 will have the problem but that's gonna hit them for more than alice probably (like symfony service definitions mainly).

Soullivaneuh:
Symfony 3.0 is now released. Did you plan a date for a new tag of this bundle? Thanks.

PR #290, opened 2015-11-16
Add quote to yaml reserved chars in fixtures.

NiR-:


Seldaek:
Would you mind updating the markdown files of the docs too? We shouldn't promote stuff that won't work..

NiR-:
@Seldaek done ;)

Soullivaneuh:
@NiR- since when using `@thing` directly cause problems with yaml? Don't have this issue on my projects.

jenkoian:
@Soullivaneuh it's to work with Symfony 3 due to https://github.com/symfony/symfony/pull/16285

Soullivaneuh:
Oh I see, thanks for the tip. :+1:

doc/relations-handling.md
```diff
@@ -15,10 +15,10 @@ Nelmio\Entity\User:
 Nelmio\Entity\Group:
     group1:
         name: Admins
-        owner: @user1
+        owner: '@user1'
 ```
 
-Alice also allows you to directly reference objects' properties using the ```@name->property``` notation.
+Alice also allows you to directly reference objects' properties using the ```'@name->property'``` notation.
```
Seldaek: Sorry but those at line 7, 35 and here should not be quoted IMO, as they are not in the Yaml. The quotes aren't part of the alice syntax.

Einenlum:
Up ?

Seldaek:
Thanks

PR #301, opened 2015-12-22
fix static constructor support

bendavies:
There is a limitation/bug in the support for static constructors, in that the static constructor required arguments needed to match the `__construct` method.

This PR fixes that use case.

### DIFF
--- a/src/Nelmio/Alice/Instances/Instantiator/Methods/ReflectionWithConstructor.php
+++ b/src/Nelmio/Alice/Instances/Instantiator/Methods/ReflectionWithConstructor.php
@@ -38,7 +38,7 @@ class ReflectionWithConstructor implements MethodInterface
      */
     public function canInstantiate(Fixture $fixture)
     {
-        $refl = new \ReflectionMethod($fixture->getClass(), '__construct');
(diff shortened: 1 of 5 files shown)


bendavies:
@Seldaek any chance of a quick review Jordi?

tshelburne:
@bendavies Makes sense to me - two things:

1. Why did you remove the arguments we had in the `User` constructor? It doesn't seem related to this PR.
1. Please create a test for your use-case that fails without your fix so we can ensure it doesn't regress in the future.

Thanks for this!

bendavies:
@tshelburne I've made this into two commits to make it a bit more obvious.

1. I've changed it up a bit. There is now a new model just for static constructor testing.
2. It's already covered. After the first commit, `testLoadCallsStaticConstructorIfProvided` will fail. The 2nd commit fixes it.

Thanks.

tshelburne:
@bendavies Thanks for this!

bendavies:
@tshelburne no problem. thanks for the merge!

bendavies:
@tshelburne any chance of a tag?

tshelburne:
@bendavies Done. Thanks again.

bendavies:
Cheers

PR #303, opened 2016-01-05
Support instanciation of object with named constructor

gquemener:
Currently having a private constructor will force instanciation using reflection (using instanciator `ReflectionWithoutConstructor`).

Current implementation only allows to use factory method when the class constructor is public.

This is not working when using the Named constructors (http://verraes.net/2014/06/named-constructors-in-php/)

This PR intends to provide the use case describing this scenario and open discussion about the best way to handle it.

So far, I've tried to manipulate the `ReflectionWithConstructor` and `ReflectionWithoutConstructor` `canInstantiate` method, but I can't make the whole test suite pass.

### DIFF
--- a/src/Nelmio/Alice/Instances/Instantiator/Methods/ReflectionWithoutConstructor.php
+++ b/src/Nelmio/Alice/Instances/Instantiator/Methods/ReflectionWithoutConstructor.php
@@ -22,6 +22,10 @@ class ReflectionWithoutConstructor implements MethodInterface
     {
         $reflConstruct = new \ReflectionMethod($fixture->getClass(), '__construct');
 
+        if (!$reflConstruct->isPublic() && '__construct' !== $fixture->getConstructorMethod()) {
+            return false;
+        }
+
         return !$reflConstruct->isPublic() || (!$fixture->shouldUseConstructor() && !version_compare(PHP_VERSION, '5.4', '<'));
     }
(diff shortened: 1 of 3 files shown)


src/Nelmio/Alice/Instances/Instantiator/Methods/ReflectionWithoutConstructor.php
```diff
@@ -22,6 +22,10 @@ public function canInstantiate(Fixture $fixture)
     {
         $reflConstruct = new \ReflectionMethod($fixture->getClass(), '__construct');
 
+        if (!$reflConstruct->isPublic() && '__construct' !== $fixture->getConstructorMethod()) {
```
tshelburne: @gquemener Let's just add this conditional bit to the line below - it's effectively the same thing, but this makes it look like there's much more change happening than there actually is. Make sense?

gquemener:
Yes it does @tshelburne, I was concerned about readability issues with a long conditionnal statement but it's not so bad in the end.

Tell me if you want me to squash the PR before merge.

tshelburne:
@gquemener Thanks!

gquemener:
Is it possible to tag a 2.1.4 with this fix please @tshelburne?

tshelburne:
@gquemener Bam! Done.

PR #305, opened 2016-01-16
added check in reference builder to enable quoting references

### LINKED ISSUE #304: Quoted references parsed as string instead of reference
Since Symfony 3.0 we can no longer use @ without it being quoted to reference a relationship, but after single quoting all my references I now get errors like 

```
Expected value of type "apiNext\Models\Merchant" for association field "apiNext\Models\Product#$merchant", got "string" instead.  
```

From a yaml file that looks like this

```yaml
product1 (extends product):
  merchant: '@merchant1'
```

where it used to work perfectly unquoted

tshelburne:
fixes #304 

@FoxxMD Can you test this and make sure it works for you?

@theofidry @Seldaek Can I get a spot-check?

### DIFF
--- a/src/Nelmio/Alice/Instances/Processor/Methods/Reference.php
+++ b/src/Nelmio/Alice/Instances/Processor/Methods/Reference.php
@@ -36,7 +36,7 @@ class Reference implements MethodInterface
      */
     public function canProcess(ProcessableInterface $processable)
     {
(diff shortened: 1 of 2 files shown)


theofidry:
:+1:

FoxxMD:
:+1: That fixed it! Thanks

PR #308, opened 2016-02-02
Small typo fix in Processors example

knugen:


knugen:
One more note regarding this actually; I just implemented basically this exact scenario with password hashing and found that it needs to happen in the `preProcess` method or it won't get persisted. If somebody confirms I will update the PR.

tshelburne:
@theofidry Can you confirm?

theofidry:
That's correct, could you move the content of `postProcess()` into `preProcess()` please @knugen?

I checked the [`ProcessorInterface` the doc](https://github.com/nelmio/alice/blob/master/src/Nelmio/Alice/ProcessorInterface.php#L24) is up to date :)

knugen:
Cool, done :)

theofidry:
:+1:

tshelburne:
Thanks guys, great catch.

PR #314, opened 2016-02-23
Fix include orders

### LINKED ISSUE #164: Suggestion on MergeInclude
```
    protected function mergeIncludeData($data, $includeData)
    {
        foreach ($includeData as $class => $fixtures) {
            if (isset($data[$class])) {
                $data[$class] = array_merge($fixtures, $data[$class]);
            } else {
                $data[$class] = $fixtures;
            }
        }

        return $data;
    }
```

Instead of the original code, here is my suggestion.
```
    protected function mergeIncludeData($data, $includeData)
    {
        foreach ($data as $class => $fixtures) {
            if (isset($includeData[$class])) {
                $includeData[$class] = array_merge($includeData[$class], $fixtures);
            } else {
                $includeData[$class] = $fixtures;
            }
        }
        return $includeData;
    }  
``` 
The included configuration will added to the beginning of the array instead of the end.
In this way, It is more connivence to write include file and reference object exits in the included files. As they will be processed first.

piribes:
* close #164

### DIFF
--- a/src/Nelmio/Alice/Fixtures/Parser/Methods/Base.php
+++ b/src/Nelmio/Alice/Fixtures/Parser/Methods/Base.php
@@ -109,11 +109,8 @@ abstract class Base implements MethodInterface
         $newData = $includeData;
 
         foreach ($data as $class => $fixtures) {
-            if (isset($newData[$class])) {
-                $newData[$class] = array_merge($newData[$class], $fixtures);
-            } else {
-                $newData[$class] = $fixtures;
-            }
+            $newData[$class] = isset($newData[$class])
+                ? array_merge($newData[$class], $fixtures) : $newData[$class] = $fixtures;
         }
 
         return $newData;
--- a/src/Nelmio/Alice/Fixtures/Parser/Methods/Base.php
+++ b/src/Nelmio/Alice/Fixtures/Parser/Methods/Base.php
@@ -102,17 +102,20 @@ abstract class Base implements MethodInterface
     /**
      * @param array $data
      * @param array $includeData
+     * @return array
      */
     protected function mergeIncludeData($data, $includeData)
     {
-        foreach ($includeData as $class => $fixtures) {
-            if (isset($data[$class])) {
-                $data[$class] = array_merge($fixtures, $data[$class]);
+        $newData = $includeData;
+
+        foreach ($data as $class => $fixtures) {
+            if (isset($newData[$class])) {
+                $newData[$class] = array_merge($newData[$class], $fixtures);
             } else {
-                $data[$class] = $fixtures;
+                $newData[$class] = $fixtures;
             }
         }
 
-        return $data;
+        return $newData;
     }
 }
--- a/tests/Nelmio/Alice/Fixtures/Parser/Methods/PhpTest.php
+++ b/tests/Nelmio/Alice/Fixtures/Parser/Methods/PhpTest.php
@@ -134,7 +134,6 @@ class PhpTest extends \PHPUnit_Framework_TestCase
                 ],
         ];
 
-        $this->assertEquals($expectedData, $data);
-        $this->assertTrue($expectedData === $data);
+        $this->assertSame($expectedData, $data);
     }
 }
--- a/tests/Nelmio/Alice/Fixtures/Parser/Methods/PhpTest.php
+++ b/tests/Nelmio/Alice/Fixtures/Parser/Methods/PhpTest.php
@@ -110,4 +110,31 @@ class PhpTest extends \PHPUnit_Framework_TestCase
         ];
         $this->assertEquals($expectedData, $data);
     }
+
+    public function testIncludeFilesOrder()
+    {
+        $data = $this->parser->parse(__DIR__.'/../../../support/fixtures/include_order.php');
+
+        $expectedData = [
+            'Foo' =>
+                [
+                    'foo' =>
+                        [
+                            'id' => 1,
+                            'text' => '<word()>',
+                        ],
+                ],
+            'Bar' =>
+                [
+                    'bar' =>
+                        [
+                            'id' => 1,
+                            'text' => '<@foo->text>'
+                        ],
+                ],
+        ];
+
+        $this->assertEquals($expectedData, $data);
+        $this->assertTrue($expectedData === $data);
+    }
 }
--- a/tests/Nelmio/Alice/Fixtures/Parser/Methods/YamlTest.php
+++ b/tests/Nelmio/Alice/Fixtures/Parser/Methods/YamlTest.php
@@ -133,9 +133,7 @@ class YamlTest extends \PHPUnit_Framework_TestCase
                 ],
         ];
 
-        $this->assertEquals($expectedData, $data);
-
-        $this->assertTrue($expectedData === $data);
+        $this->assertSame($expectedData, $data);
     }
 
     public function testParametersNotReturnedInData()
--- a/tests/Nelmio/Alice/Fixtures/Parser/Methods/YamlTest.php
+++ b/tests/Nelmio/Alice/Fixtures/Parser/Methods/YamlTest.php
@@ -110,6 +110,34 @@ class YamlTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals($expectedData, $data);
     }
 
+    public function testIncludeFilesOrder()
+    {
+        $data = $this->parser->parse(__DIR__ . '/../../../support/fixtures/include_order.yml');
+
+        $expectedData = [
+            'Foo' =>
+                [
+                    'foo' =>
+                        [
+                            'id' => 1,
+                            'text' => '<word()>',
+                        ],
+                ],
+            'Bar' =>
+                [
+                    'bar' =>
+                        [
+                            'id' => 1,
+                            'text' => '<@foo->text>'
+                        ],
+                ],
+        ];
+
+        $this->assertEquals($expectedData, $data);
+
+        $this->assertTrue($expectedData === $data);
+    }
+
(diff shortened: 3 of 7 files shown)


src/Nelmio/Alice/Fixtures/Parser/Methods/Base.php
```diff
@@ -102,17 +102,18 @@ protected function processIncludes($data, $filename)
...
+            if (isset($includeData[$class])) {
+                $includeData[$class] = array_merge($includeData[$class], $fixtures);
             } else {
-                $data[$class] = $fixtures;
+                $includeData[$class] = $fixtures;
             }
         }
 
-        return $data;
+        return $includeData;
```
tshelburne: @piribes I feel like changing `$includeData` makes the intention here less clear, since conceptually `$includeData` should be appended to `$data`. Instead, let's just create a new variable, `$newData` or something, and return it. Does that make sense?

piribes: It makes sense to me, i have updated the code.

src/Nelmio/Alice/Fixtures/Parser/Methods/Base.php
```diff
@@ -102,17 +102,20 @@ protected function processIncludes($data, $filename)
...
      * @param array $data
      * @param array $includeData
+     * @return array
      */
     protected function mergeIncludeData($data, $includeData)
     {
-        foreach ($includeData as $class => $fixtures) {
-            if (isset($data[$class])) {
-                $data[$class] = array_merge($fixtures, $data[$class]);
+        $newData = $includeData;
```
theofidry: `$newData` is unneeded, you can use `$includeData` right away

src/Nelmio/Alice/Fixtures/Parser/Methods/Base.php
```diff
@@ -102,17 +102,20 @@ protected function processIncludes($data, $filename)
...
+
+        foreach ($data as $class => $fixtures) {
+            if (isset($newData[$class])) {
```
theofidry: this `if/else` can be avoid by using a ternary operator:

```php
$newData[$class] = (isset($newData[$class])) ? array_merge($newData[$class], $fixtures) : $fixtures;
```

tests/Nelmio/Alice/Fixtures/Parser/Methods/PhpTest.php
```diff
@@ -110,4 +110,31 @@ public function testIncludeFiles()
...
+                    'bar' =>
+                        [
+                            'id' => 1,
+                            'text' => '<@foo->text>'
+                        ],
+                ],
+        ];
+
+        $this->assertEquals($expectedData, $data);
+        $this->assertTrue($expectedData === $data);
```
theofidry: Why this additional assertion?

tests/Nelmio/Alice/Fixtures/Parser/Methods/YamlTest.php
```diff
@@ -110,6 +110,34 @@ public function testIncludeFiles()
...
+                        [
+                            'id' => 1,
+                            'text' => '<@foo->text>'
+                        ],
+                ],
+        ];
+
+        $this->assertEquals($expectedData, $data);
+
+        $this->assertTrue($expectedData === $data);
```
theofidry: same as https://github.com/nelmio/alice/pull/314/files#r53787486

tests/Nelmio/Alice/Fixtures/Parser/Methods/PhpTest.php
piribes: PHPUnit assertsEquals() compare $expectedData and $data using == operator.
This does not ensure that the data is in correct order, but it is useful to get the differences in output.

src/Nelmio/Alice/Fixtures/Parser/Methods/Base.php
piribes: @theofidry i created $newData instead of reusing $includeData just for code clarity, as stated here by @tshelburne 
If you think it is incorrect, there's no problem, we can rollback whenever we want, of course :)

tests/Nelmio/Alice/Fixtures/Parser/Methods/PhpTest.php
theofidry: >but it is useful to get the differences in output.

What do you mean? In this case they do exactly the same thing as they are arrays and not objects, the only difference is that `===` will make sure they are of the same type, hence `array` here. As you're giving the expected result (which is a non-empty array), this check is unneeded.

src/Nelmio/Alice/Fixtures/Parser/Methods/Base.php
theofidry: To me it's just that in this case if the name is not clear enough it should be changed in the signature :)

That being said I'm saying that because I find this line useless that's it. If you or @tshelburne find it more clear to do it this way I'm fine with it.

tests/Nelmio/Alice/Fixtures/Parser/Methods/PhpTest.php
piribes: I'll try to explain it better.

assertEquals() will print diff-style message when values of $expectedData and $data are not the same, helping you to find out why the test has failed.

Here is an example of what you'll get if for some reason the test subject is different from what expected:

```
1) Nelmio\Alice\Fixtures\Parser\Methods\YamlTest::testIncludeFilesOrder
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     'Foo' => Array (
         'foo' => Array (
-            'id' => 2
+            'id' => 1
             'text' => '<word()>'
         )
     )
     'Bar' => Array (...)
 )
```
As assertEquals() will use == operator to compare the arrays, it will not check if the order is the same.

assertTrue() instead, will ensure that sorting is preserved, but unlike assertEquals(), it will not print a diff-style message to help you.

Another way to write it with just one assert, could be:
```
$this->assertSame($expectedData, $data)
```

This will print a slightly different message from those printed by assertEquals() in case of failure. Here is an example:

```
1) Nelmio\Alice\Fixtures\Parser\Methods\YamlTest::testIncludeFilesOrder
Failed asserting that Array &0 (
    'Foo' => Array &1 (
        'foo' => Array &2 (
            'id' => 1
            'text' => '<word()>'
        )
    )
    'Bar' => Array &3 (
        'bar' => Array &4 (
            'id' => 1
            'text' => '<@foo->text>'
        )
    )
) is identical to Array &0 (
    'Foo' => Array &1 (
        'foo' => Array &2 (
            'id' => 2
            'text' => '<word()>'
        )
    )
    'Bar' => Array &3 (
        'bar' => Array &4 (
            'id' => 1
            'text' => '<@foo->text>'
        )
    )
).
```

theofidry: >it will not check if the order is the same.

But `[1,2] == [2,1]` will return `false`, regardless of if you are using `==` or `===`. That's why I'm a bit lost with your explanation.

src/Nelmio/Alice/Fixtures/Parser/Methods/Base.php
tshelburne: This is per my request earlier - I feel like this variable name makes the intention much more clear.

src/Nelmio/Alice/Fixtures/Parser/Methods/Base.php
tshelburne: I can go either way on this.

tests/Nelmio/Alice/Fixtures/Parser/Methods/PhpTest.php
piribes: You are right, but in this example you are comparing [0=>1,1=>2] with [0=>2,1=>1], hence the == operator return false because $first[0] != $second[0].

Instead, if you compare [1=>1,2=>2] and [2=>2,1=>1] the == operator return true because $first[1] == $second[1] and $first[2] == $second[2] no matter in what order they are.

On the other hand, if you use === operator you'll get false, because you are asking to compare also the orders.

[Here](http://php.net/manual/en/language.operators.array.php) you can find the documentation of array operators.

theofidry: Ok fair. But do we have to test the order actually?

piribes: Sure, this is exactly what we want to test.

The order returned by parse() must be exactly the same as in $expectedData, otherwise we have lost the "include" order as defined in the parsed fixture file.
Please refer to issue #314 for further details.

Now that this is clarified, what are we going to do ?
Do you think it would be better to use assertEquals() + assertTrue() or assertSame() only ? :)

theofidry: >Do you think it would be better to use assertEquals() + assertTrue() or assertSame() only ?

I think we can go for the later. If we want more details on the error for debug purposes we can change it locally for debugging :) That being said I don't have a strong opinion on that one.

piribes: Agree, assertSame() also better explain the intent, compared to assertTrue() solution.

piribes:
Any chance ?

theofidry:
:+1: As far as I can tell there is no BC break so can be released in a patch :)

tshelburne:
Makes sense to me - @piribes thanks!

piribes:
my pleasure

PR #323, opened 2016-04-06
Add support for snake_care properties in fixtures.

hboomsma:
When for a property a setter exists Alice will use that setter to populate the value. Now properties that use snake case are also transformed into their corresponding setters. For example:

```yaml
Nelmio\Alice\support\models\User:
    user{1..10}:
        family_name: <lastName()>
```
    
Will use the setter setFamilyName if available and otherwise fill the property without setter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### 2.x (WIP)
+
+  * Added support for snake_case properties (#323)
+
 ### 2.1.2 (2015-12-10)
 
   * Bug fix - private properties were populatable directly on the instance, but not private properties of parent classes. Although this is an antipattern, if we're allowing it for the instance we should allow it up the chain.
--- a/src/Nelmio/Alice/Instances/Populator/Methods/Direct.php
+++ b/src/Nelmio/Alice/Instances/Populator/Methods/Direct.php
@@ -31,7 +31,7 @@ class Direct implements MethodInterface
      */
     public function canSet(Fixture $fixture, $object, $property, $value)
     {
-        return method_exists($object, $this->setterFor($property));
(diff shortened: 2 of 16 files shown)


reinfi:
A good codingstyle would have the setters also in snake_case, not in camelCase.
So what's the case you need this?

hboomsma:
Having a PSR2 compatible code style with snake case properties. And several MLOC, I would rather not change. Already implemented this as extension in the company, just trying to give something back by making Alice more codestyle agnostic.

reinfi:
Do I understand you right that you have snake_case properties and camelCase setters?

hboomsma:
correct.

tshelburne:
I think I'm good for this, since it doesn't break BC - @Seldaek any comments?

Seldaek:
Seems safe enough as long as people use camelCased method names, but strictly speaking it does break BC as if you have setfoo_bar it'll now call setfoobar. I doubt it'll really be an issue in practice though.

tshelburne:
@hboomsma I think @Seldaek makes a good point - why don't we just check for both methods and call whichever exists?

hboomsma:
@tshelburne I agree, added it in.

theofidry:
@hboomsma I've made a PR with a few changes.

If you have a property `full_name`, it will look in order:

- `setfullname()`
- `setfull_name()`
- `set_full_name()`
- `set_fullname()`

If it is a normal property, e.g. `name`:

- `setname()`
- `set_name()`

It also avoids to introduce a BC @Seldaek mentioned.

theofidry:
Thanks very much for the work @hboomsma :)

PR #326, opened 2016-04-24
[v3] Rework instantiators

theofidry:
@tshelburne here's the first part I feel like re-working: Instantiators.

The first main changes I would like to introduce are:

- **Have an interface for Instantiator**. It gives more flexibility, and will allow us to have final classes behind. Final classes relying on interfaces provide the benefit to avoid BC break later on
- **Allow failure by throwing an exception**

Internal change:

- remove `Methods` for chainable instantiator instead.
- make classes final by default
- make classes immutable by default

### Point to discuss besides the review

**PHP version**: the current code assumes PHP7. I can easily understand it being a bit too forward. However realistically, v3 is not going to come out before months so PHP7 will already be more adopted at that time. Besides, v2 has been out for over a year already and has not know lots of new features. So I would say PHP7 helps us a lot for maintenance and user who cannot upgrade can always use v2.

**Migration strategy**: ~~for now I'm not removing the old code, just deprecating it. So we could still tag it under v2 and do the last move for v3. This way, user would be noticed of the changes as they go on even if the upgrade is unlikely to be smooth.~~

For now, at least for this PR, I'm trying to add the new pieces without breaking the whole test suite.

### Notes

I'm trying to make the tests and interfaces as self-explanatory as possible, if there is a question, this is what should be checked first.


theofidry:
@tshelburne updated. I think there is a lot to discuss about especially as it's the first change.

theofidry:
EDIT: what's missing from the PR:

- `ReflectionWithoutConstructor`: not really sure to understand the use case
- `Unserialize`: not really sure to understand it either; looks like a hack for <PHP5.4 so I suggest to kill it.
- `ReflectionWithConstructor`: waiting to review/work `Processor` before tackling this one

theofidry:
@tshelburne would it be possible to have a quick review? Or rather at least agree on the PHP version used as it's really the blocking part for me :)

src/Nelmio/Alice/Exception/Instantiator/RuntimeException.php
```diff
@@ -0,0 +1,20 @@
...
+ *  
+ *  (c) Nelmio <hello@nelm.io>
+ *  
+ *  For the full copyright and license information, please view the LICENSE
+ *  file that was distributed with this source code.
+ */
+
+namespace Nelmio\Alice\Exception\Instantiator;
+
+/**
```
theofidry: TBR

src/Nelmio/Alice/Exception/Instantiator/ExceptionInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ *  For the full copyright and license information, please view the LICENSE
+ *  file that was distributed with this source code.
+ */
+
+namespace Nelmio\Alice\Exception\Instantiator;
+
+/**
+ * @author Théo FIDRY <theo.fidry@gmail.com>
+ */
+interface ExceptionInterface extends \Throwable
```
theofidry: The name might be confusing:

- When using it without aliases, you have `ExceptionInterface`, `InstantiatorExceptionInterface` would maybe be clearer
- It's weird to have `Exception` when technically speaking it's a `Throwable`, should be renamed to reflect on that?

src/Nelmio/Alice/Instances/Instantiator/Chainable/EmptyConstructorInstantiator.php
```diff
@@ -0,0 +1,43 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\Alice\Instances\Instantiator\Chainable;
+
+use Nelmio\Alice\Fixtures\Fixture;
+use Nelmio\Alice\Instances\Instantiator\ChainableInstantiatorInterface;
+
+final class EmptyConstructorInstantiator implements ChainableInstantiatorInterface
```
theofidry: rework of `Instances\Instantiator\Methods\EmptyConstructor`

src/Nelmio/Alice/Instances/Instantiator/Chainable/EmptyConstructorInstantiator.php
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * @inheritdoc
+     */
+    public function canInstantiate(Fixture $fixture): bool
+    {
+        try {
+            $reflectionMethod = new \ReflectionMethod($fixture->getClass(), '__construct');
+
+            return 0 === $reflectionMethod->getNumberOfRequiredParameters();
+        } catch (\ReflectionException $exception) {
```
theofidry: notable change with the original class: the `try/catch` block

src/Nelmio/Alice/Instances/Instantiator/ChainableInstantiatorInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\Alice\Instances\Instantiator;
+
+use Nelmio\Alice\Fixtures\Fixture;
+use Nelmio\Alice\Instances\InstantiatorInterface;
+
+/**
+ * @author Théo FIDRY <theo.fidry@gmail.com>
```
theofidry: kill `@author`

src/Nelmio/Alice/Instances/Instantiator/InstantiatorRegistry.php
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+namespace Nelmio\Alice\Instances\Instantiator;
+
+use Nelmio\Alice\Exception\Instantiator\RuntimeException;
+use Nelmio\Alice\Fixtures\Fixture;
+use Nelmio\Alice\Instances\InstantiatorInterface;
+
+/**
+ * @author Théo FIDRY <theo.fidry@gmail.com>
```
theofidry: kill `@author`

src/Nelmio/Alice/Instances/Instantiator/InstantiatorRegistry.php
```diff
@@ -0,0 +1,68 @@
...
+ */
+final class InstantiatorRegistry implements InstantiatorInterface
```
theofidry: rework of the old `Instantiator` class, there is little differences except that it's immutable

src/Nelmio/Alice/Instances/InstantiatorInterface.php
```diff
@@ -0,0 +1,32 @@
...
+interface InstantiatorInterface
+{
+    /**
+     * Creates and returns an instance of the class described by the given fixture.
+     *
+     * @param  Fixture $fixture
+     *
+     * @throws ExceptionInterface
+     *
+     * @return mixed
```
theofidry: maybe should be changed to return `object` instead?

theofidry:
Only just though about that but maybe HautelookAliceBundle contributors want to participate as well? /cc @Soullivaneuh @vincentchalamon

tests/PhpUnit/Assert.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+namespace PhpUnit;
+
+use PHPUnit_Framework_TestCase as PhpUnit;
+
+/**
+ * @author Théo FIDRY <theo.fidry@gmail.com>
+ */
+final class Assert extends \PHPUnit_Framework_TestCase
```
theofidry: should be renamed `PhpUnit` or `Assertor`, `Assert::assertIsA()` doesn't look good

Soullivaneuh:
> for now I'm not removing the old code, just deprecating it. So we could still tag it under v2 and do the last move for v3.

[...]

> the current code assumes PHP7. I can easily understand it being a bit too forward. However realistically, v3 is not going to come out before months so PHP7 will already be more adopted at that time. 

Well, deprecation things should be on a new minor IMHO and not a on major with PHP version support dropping.

theofidry:
@Soullivaneuh I think deprecations could be added to `1.x` and not worry about that in `master`, and have a good changelog.

src/Nelmio/Alice/Fixtures/Fixture.php
```diff
@@ -342,12 +342,16 @@ protected function getConstructor()
...
-    // array('foo', 'bar') => new $fixture->getClass()('foo', 'bar')
-    // array('foo' => array('bar')) => $fixture->getClass()::foo('bar')
-    //
+    /**
+     * Sequential arrays call the constructor, hashes call a static method
+     *
+     * array('foo', 'bar') => new $fixture->getClass()('foo', 'bar')
+     * array('foo' => array('bar')) => $fixture->getClass()::foo('bar')
+     *
+     * @throws \UnexpectedValueException
```
theofidry: should be ported to 2.x as well

theofidry:
@tshelburne updated, tests should pass (they do locally) on Travis once #335 and #336 will be merged.

PR #331, opened 2016-05-03
Deprecate misusage of range

### LINKED ISSUE #329: [v2] Deprecate range
As of now, the [current regex](https://github.com/nelmio/alice/blob/3bf8e6f9cabb4f7d06d0c0d8f715ceb4b41d19c9/src/Nelmio/Alice/Fixtures/Builder/Methods/RangeName.php#L25) (`#\{([0-9]+)\.\.(\.?)([0-9]+)\}#i`) allows both `user{1..10}` and `user{1...10}`. This makes it error prone which can be nice.

However, I think this can confuse the user as well (does it work well? Is it by design I as can't see anything in the doc about that?) and also if we support that I see no reason to not support `user{0......10}` for example.

Instead what about removing this behaviour in v3 and deprecating it in v2?

theofidry:
Closes #329

Proposed changes:

- Add deprecation warning (no test for it)
- Added tests for `RangeName`
- Fixed `valueForCurrent` to be a string regardless of the index (previously could be `0` and then `'1'`...

~~I must say I also find it weird to have `user_{1...10}` creating the fixtures `user_1`, ... `user_9` where I would expect it to go up to `user_10`.~~

Looks like it's a misunderstanding of my part: `user{0...10}` is creating 10 users, not 9, whereas `user{0..10}` creates 10 users as expected...

### DIFF
--- a/src/Nelmio/Alice/Fixtures/Builder/Methods/RangeName.php
+++ b/src/Nelmio/Alice/Fixtures/Builder/Methods/RangeName.php
@@ -43,7 +43,7 @@ class RangeName implements MethodInterface
         $fixtures = [];
 
         $from = $this->matches[1];
-        $to = $this->matches[2] - 1;
+        $to = $this->matches[2];
         if ($from > $to) {
             list($to, $from) = [$from, $to];
         }
--- a/src/Nelmio/Alice/Fixtures/Builder/Methods/RangeName.php
+++ b/src/Nelmio/Alice/Fixtures/Builder/Methods/RangeName.php
@@ -22,17 +22,18 @@ class RangeName implements MethodInterface
      */
     public function canBuild($name)
     {
-        if (1 === preg_match('/\{([0-9]+)\.{3,}([0-9]+)\}/', $name, $this->matches)) {
+        if (1 === preg_match('/\{([0-9]+)(\.{3})([0-9]+)\}/', $name, $this->matches)) {
             @trigger_error(
                 'Ranged name should follow the pattern "name{X..Y}". Using "name{X...Y} instead is now deprecated and '
-                .'will be removed in 3.0',
+                .'will be removed in 3.0. Please mind the change of behavior: "user{0..10}" is creating 11 users '
+                .'whereas "user{0...10}" is creating 10',
                 E_USER_DEPRECATED
             );
 
             return true;
         }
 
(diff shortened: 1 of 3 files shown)


src/Nelmio/Alice/Fixtures/Builder/Methods/RangeName.php
```diff
@@ -22,7 +22,16 @@ class RangeName implements MethodInterface
      */
     public function canBuild($name)
     {
-        return 1 === preg_match('#\{([0-9]+)\.\.(\.?)([0-9]+)\}#i', $name, $this->matches);
+        if (1 === preg_match('/\{([0-9]+)\.{3,}([0-9]+)\}/', $name, $this->matches)) {
+            @trigger_error(
+                'Ranged name should follow the pattern "name{X..Y}". Using "name{X...Y} instead is now deprecated and will be removed in 3.0',
+                E_NOTICE
```
Soullivaneuh: `E_USER_DEPRECATED`

src/Nelmio/Alice/Fixtures/Builder/Methods/RangeName.php
```diff
@@ -22,7 +22,16 @@ class RangeName implements MethodInterface
      */
     public function canBuild($name)
     {
-        return 1 === preg_match('#\{([0-9]+)\.\.(\.?)([0-9]+)\}#i', $name, $this->matches);
+        if (1 === preg_match('/\{([0-9]+)\.{3,}([0-9]+)\}/', $name, $this->matches)) {
+            @trigger_error(
+                'Ranged name should follow the pattern "name{X..Y}". Using "name{X...Y} instead is now deprecated and will be removed in 3.0',
```
Soullivaneuh: This line is too long, could you please lime break?

tests/Nelmio/Alice/Fixtures/Builder/Methods/RangeNameTest.php
```diff
@@ -0,0 +1,134 @@
...
+            'with no upper bound' => ['user_{0..}', false],
+        ];
+    }
+
+    public function provideData()
+    {
+        $class = 'Dummy';
+        $specs= [];
+
+        yield [
```
Soullivaneuh: yield is only available since php 5.5.

theofidry:
Updated to reflect on the `..`, `...` difference

PR #340, opened 2016-05-07
[v2] Review Parsers

theofidry:
Proposed changes:

- Minor changes in the PHPDoc
  - fix type/formating
  - add `@throws` tag when an exception is thrown
- Rework tests
  - add new tests methods
  - moved fixtures files and tried to rework them to put only what is under test and avoid any confusion


src/Nelmio/Alice/Fixtures/Parser/Methods/Base.php
```diff
@@ -103,17 +109,29 @@ protected function processIncludes($data, $filename)
...
+        }
 
+        if (false === is_array($includeData)) {
+            return $data;
+        }
+
+        $newData = $includeData;
         foreach ($data as $class => $fixtures) {
-            $newData[$class] = isset($newData[$class])
-                ? array_merge($newData[$class], $fixtures) : $newData[$class] = $fixtures;
```
theofidry: needs to check the result when parameters are being used as well

PR #347, opened 2016-05-13
add reference range name builder

reinfi:
This builder would help to create fixtures for OneToOne relations within doctrine or other ORMs.

You could reference objects created earlier in your fixtures to create the same amount of instances.
You're free to select just one object or the whole list matched by a mask. The builder then creates as many instances as matched by your mask.

The only condition is that your fixtures must be declared within different files.

I saw this pull request #236. In a way it gives the opportunity to add own builders but in my case I need the object reference. Another idea could be to implemnent an ObjectsAwareInterface which sets the current objects to a builder if required.


reinfi:
Still no idea why the build fails, need to see if I can fix it tomorrow.

src/Nelmio/Alice/Fixtures/Builder/Methods/ReferenceRangeName.php
```diff
@@ -0,0 +1,78 @@
...
+                sprintf('Instance %s is not defined!', $this->matches[1])
+            );
+        }
+        $currentName = str_replace($this->matches[0], $referenceName, $name);
+
+        $fixtures[] = new Fixture($class, $currentName, $spec, $currentValue);
+
+        return $fixtures;
+    }
+}
```
theofidry: needs an empty line at the end of the file

src/Nelmio/Alice/Fixtures/Builder/Methods/ReferenceRangeName.php
```diff
@@ -0,0 +1,78 @@
...
+    public function build($class, $name, array $spec)
+    {
+        $fixtures = [];
+
+        // could be 'car1' from engine_{@car1}
+        $referenceName = $this->matches[1];
+        $referenceAll = '*' === $this->matches[2];
+
+        if ($referenceAll) {
+            $keys = $this->objects->getKeysByMask($referenceName.".+");
```
theofidry: let's put L48-L63 in a separate private function, e.g. `buildAll()`

src/Nelmio/Alice/Fixtures/Builder/Methods/ReferenceRangeName.php
```diff
@@ -0,0 +1,78 @@
...
+
+            if (count($keys) === 0) {
```
theofidry: yoda condition

src/Nelmio/Alice/Fixtures/Builder/Methods/ReferenceRangeName.php
```diff
@@ -0,0 +1,78 @@
...
+                $currentName = str_replace($this->matches[0], $key, $name);
+
+                $fixtures[] = new Fixture($class, $currentName, $spec, $instance);
+            }
+
+            return $fixtures;
+        }
+
+        $currentValue = $this->objects->get($this->matches[1]);
+        if (is_null($currentValue)) {
```
theofidry: `null !== $currentValue`

tests/Nelmio/Alice/ReferenceRangeNameTest.php
```diff
@@ -0,0 +1,104 @@
...
+namespace Nelmio\Alice;
+
+class ReferenceRangeNameTest extends \PHPUnit_Framework_TestCase
+{
+    const USER = 'Nelmio\Alice\support\models\User';
+    const GROUP = 'Nelmio\Alice\support\models\Group';
+    const CONTACT = 'Nelmio\Alice\support\models\Contact';
+
+    /**
+     * @test
```
theofidry: let's keep consistent with the rest of the project and prefix the test methods with `test` instead of using the `@test` annotation

tests/Nelmio/Alice/ReferenceRangeNameTest.php
```diff
@@ -0,0 +1,104 @@
...
+     * @expectedException \UnexpectedValueException
+     */
+    public function loadFixturesByReferenceNotFound()
```
theofidry: `testThrowExceptionWhenReferencesAreNotFound()`

tests/Nelmio/Alice/ReferenceRangeNameTest.php
```diff
@@ -0,0 +1,104 @@
...
+        $files = [
+            __DIR__ . '/support/fixtures/reference_range_1.yml',
+            __DIR__ . '/support/fixtures/reference_range_2.yml',
+        ];
+        $objects = Fixtures::load($files, $managerMock, [ 'providers' => [ $this ] ]);
+
+        $this->assertCount(7, $objects);
+
+        $groupReference = $objects['group_1_user1'];
+        $this->assertInstanceOf(self::GROUP, $groupReference);
```
theofidry: besides testing that the constructed objects are of the right instance, it would be good to test the result value of `<current()>` for the `Group` entities

theofidry:
It *is* weird, they do pass locally when I pull out your branch...

tests/Nelmio/Alice/ReferenceRangeNameTest.php
theofidry: thought about it: must also be tested when the referenced ranged is defined by a range list (`user_{alice, bob}`).

Also worth to add a test with a third reference list defined in the same fashion

reinfi: I already check the value of <current()>:

$this->assertInstanceOf(self::USER, $groupReference->getOwner());

Or what do you mean? 
I will add another test with a range list.

theofidry: but you are just testing that `$groupReference->getOwner()` returns a user, in this case it's not enough: it's must be a very specific user right?.

reinfi: you're right, I only made those checks later on. Added it.

theofidry:
@reinfi any leads on what is going wrong? Sorry I was a bit busy with #355...

reinfi:
Finally I found it. I merged the latest changes in branch 2.x, it seems travis does this automatically. There were some changes so I fixed the test and the new builder to work with those changes.

theofidry:
Nice! I'm waiting for #355 to be merged and then I'll merge yours :)

theofidry:
@reinfi #355 as been merged! Could you please rebase your work? I hope you won't get too many conflicts...

reinfi:
Sorry @theofidry, this got totally out of my mind. I merged the latest 2.x branch now. What else is missing?

theofidry:
I think there was only the tests to fix left :)

I'm gonna take a new look at it and to a PR against master as well to ensure this feature is on both 2.x and 3.x. Thanks

theofidry:
Closed via #693, it has been released in 2.3.0 :)

PR #355, opened 2016-05-14
Harden parameters and references

### LINKED ISSUE #284: Object IDs with slashes
With the following fixtures file:

```yml
AppBundle\Entity\PuppetPackage:
    puppet_package_app-admin/puppet:
        name: app-admin/puppet
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_dev-db/mysql:
        name: dev-db/mysql
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_dev-db/phpmyadmin:
        name: dev-db/phpmyadmin
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_dev-db/postgresql-server:
        name: dev-db/postgresql-server
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_dev-lang/php:
        name: dev-lang/php
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_kernel:
        name: kernel
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_mail-filter/spamassassin:
        name: mail-filter/spamassassin
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_mail-mta/postfix:
        name: mail-mta/postfix
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_net-dns/bind:
        name: net-dns/bind
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_net-misc/openssh:
        name: net-misc/openssh
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_sys-apps/baselayout:
        name: sys-apps/baselayout
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_sys-apps/openrc:
        name: sys-apps/openrc
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_sys-apps/portage:
        name: sys-apps/portage
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_sys-boot/grub:
        name: sys-boot/grub
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_www-servers/apache:
        name: www-servers/apache
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0
    puppet_package_www-servers/nginx:
        name: www-servers/nginx
        availableVersions: [1.0.0, 1.0.1, 2.0.0]
        defaultVersion: 2.0.0

AppBundle\Entity\PuppetClass:
    puppet_class_apache:
        name: Apache
        packages:
            - @puppet_package_www-servers/apache
    puppet_class_php:
        name: PHP
        packages:
            - @puppet_package_dev-lang/php
    puppet_class_mysql:
        name: Mysql
        packages:
            - @puppet_package_dev-db/mysql
            - @puppet_package_dev-db/phpmyadmin
```

I got errors like:

```
Entity with Id @puppet_package_www-servers/apache and Class AppBundle\Entity\PuppetPackage not found
```

If I use `@kernel` package reference, it works. I think there is an issue with IDs containing a `/`.

I dumped IDs on `Loader::populateObjects()` and I get corrects one.

Is that a bug?

### LINKED ISSUE #358: Incorrect number with ranged fixtures
If I have:

```yaml
User:
  user_{2...0}:
    # ...
```

It will generate... 4 users.

First this problem affects the deprecated range usage (using more than two dots, cf. #329). This also raise the question of using negative numbers in range. This was kind of possible before, but likely to fail when using the generated name in references. But as `-` is likely to be an issue in a reference, I'm for completely disabling this possibility.

theofidry:
Closes #284
Closes #358

Proposed changes:

- Fixture references should be lowercase and only be composed of letters, digits, periods (`.`), underscores (`_`) and slashes (`/`). The current policy is to deprecate it, although I expect one case or two to no longer work; but those should be very weird edge cases that are more bugs than anything.
- Rework `Parameterized` processor to process only `Processable`s which holds a parameter
- Add **LOADS** of tests
- Add some warning in the doc regarding fixture references names and parameters

/cc @Soullivaneuh for #284.


src/Nelmio/Alice/Fixtures/Builder/Builder.php
```diff
@@ -87,6 +89,16 @@ public function build($class, $name, array $spec)
                 return $fixtures;
             }
         }
+
+        @trigger_error(
```
theofidry: Needs to be added to the changelog

doc/relations-handling.md
```diff
@@ -18,6 +18,10 @@ Nelmio\Entity\Group:
         owner: '@user1'
 ```
 
+**Warning:** References (e.g. here `group1`) should always be lowercase and not containing any of the following special
+characters: `{`, `}`, `-`, `>`, `'`. They all are reserved characters, they may be used but will have a special meaning
```
theofidry: To be changed: can be composed of letters + digits + `.`, `_`, `/` and must contain at least one letter

tests/Nelmio/Alice/Fixtures/Builder/Reference.php
```diff
@@ -0,0 +1,232 @@
...
+
+namespace Nelmio\Alice\Fixtures\Builder\Methods;
+
+use Nelmio\Alice\Fixtures\Fixture;
+
+/**
+ * List a list of fixture reference, with the expected result. The list is flagged to be able to easily extract some
+ * parts of it. This class aims at centralising all the possible fixture tested to avoid too much duplication.
+ */
+final class Reference
```
theofidry: @tshelburne I know you're very busy, but I would like you to review that one as it's a fairly big changed. The idea in this class is to provide a full list of possible values which will be used for tests. I'll add comments below to explain the format. There **is BC breaks**. I avoided them at all cost in favour of deprecations, but there is a few of them that are, IMO more bugs. Still, it's my opinion and I may have got a few things wrong, so please do check that one in details.

tests/Nelmio/Alice/Fixtures/Builder/Reference.php
```diff
@@ -0,0 +1,232 @@
...
+{
+    private static $references;
+
+    private function __construct()
+    {
+        self::$references = [
+            'simple' => [
```
theofidry: the first key is just to categorise the whole block

tests/Nelmio/Alice/Fixtures/Builder/Reference.php
```diff
@@ -0,0 +1,232 @@
...
+                'simple' => [
```
theofidry: the format is:

```php
'title of the set of data' => [
  'name of the fixture',
  [ expected fixtures ], // null if no builder/method can build it
]
```

tests/Nelmio/Alice/Fixtures/LoaderTest.php
```diff
@@ -897,20 +887,26 @@ public function testLoadCreatesEnumsOfObjects()
     {
         $res = $this->loadData([
             self::USER => [
-                'user_{alice, bob, foo bar}' => [
```
theofidry: non acceptable BC break, should be fixed

tests/Nelmio/Alice/Fixtures/Builder/Reference.php
theofidry: Notable changes with the system prior to that PR:

- `user_{alice, bob,}` previously was building a reference named `user_{alice, bob,}`. Now builds that as a list, i.e. result in `user_alice` and `user_bob`. A deprecation warning is also thrown to warn the user that the list is poorly formatted and an exception will be thrown in v3.
- `user_{, alice, bob}`: same as previous case.
- `user_{0..2}`: value for `<current()>` were respectively `'0'`, `1`, `2`. Now are all strings as states the phpdoc. Changed in #339.
- `user_{0....2}`: was generating only one fixture named `user_{0....2}`; now is equivalent to `user_{0...2}`
- `user_{2...0}`: as reported in #358 was generating 4 fixtures... now is equivalent to `0...2` which result in `user_0`, `user_1` and `user_2`
- `user_{2...2}`: was generating two fixtures `user_1` and `user_3`, now doesn't build any (the segment is `[2;2[` so contains no element)
- `user_{0.2}`: was generating a fixture named `user_{0.2}`. No doesn't generate any.
- `user_{2..}`: was generating a fixture named `user_{2..}`. No doesn't generate any.
- `user_{-1..2}`: was generating a fixture named `user_-1..2`. No doesn't generate any. Same goes for all ranges containing a negative number

You can guess that they required to drastically change the behaviour of the Builder methods, which *is* a BC break. However as it's a fairly low level thing and this is necessary to fix those bugs, I reckon it can be changed?

Makes me thing that those Methods should be annotated as internal.

doc/relations-handling.md
tshelburne: Sounds good.

src/Nelmio/Alice/Fixtures/Builder/Methods/ListName.php
```diff
@@ -25,21 +25,57 @@ class ListName implements MethodInterface
      */
     public function canBuild($name)
     {
-        return 1 === preg_match('#\{([^,]+(\s*,\s*[^,]+)*)\}#', $name, $this->matches);
+        if (1 !== preg_match('/\{(?<content>.*)\}/', $name, $this->matches)) {
```
tshelburne: I feel like too much is happening here - what about just breaking this class into two?

tshelburne:
Leaving this open to review in the next couple days. Didn't realize how big it was going to be.

matt-usurp:
Is there any reason that the fixture names are required to be lowercase now? In projects at work we have data fixtures arranged in directories that match the directory structure of our entities, the references also follow this directory structure.

For example
```YAML
# src/SomeBundle/DataFixtures/ORM/SomeDirectory/SomeClass.yml
# src/SomeBundle/Entity/SomeDirectory/SomeClass.php
SomeBundle\Entity\SomeDirectory\SomeClass:
  SomeDirectory_SomeClass.single_fixture:
    name: hello
```

This allows us to instantly have a none-clashing reference for all our fixtures and be able to locate the relevant `.yml` file that it lives in. Removing the ability to add uppercase characters means the references also increase in length as we need to start adding additional characters. `some-directroy_some-class.single_fixture` for example.

We also have this case where its nice to list all the names in the fixtures that are canonical and generate slugs or alias based on those, in this case we would need to reverse engineer a list entry to a readable name. Lets take for example (I am not sure if punctuation is valid but the other examples are valid still)

```YAML
Class:
  "reference_{Alice,White Rabbit,Mad Hatter,Alice's Cat}":
    name: <current()>
    alias: <(someUrlFriendlyFunction($name))>
```

I guess if it was enforced that lowercase is mean't to happen I would make a template and have many more lines and remove all instances of the `{list}` syntax.

Just my thoughts.

theofidry:
Thanks for sharing your thoughts @matt-usurp.

My reasons for those changes:

- I find very error prone to have `@user` and `@User` that would give two different results
- Special characters are problematic as some are already used: `{`, `}`, `-`, `(`, `)`, `>` and complicates a lot of other regex, which makes it harder to maintain and ensure they work correctly
- Not every regex are case-sensitive, which means there is already potential bugs when relying on it
- We have absolutely no test for such cases which means we can't guarantee the result
- It's common for service declaration to have lowercase service names with a limited set of special characters (usually `.`, `_` and `-`).

The rationale behind is that right now, there is not much checks done. This is cool as it offers great flexibility. The issue is that it as a lot, lot more testing. The current tests are good enough for most of the common use cases but they often miss dark edge cases. Having "weird" references is very far from being tested thoroughly if at all. Hence my intend to offer less flexibility on this side to ease maintenance and develop feature on our side and offer better reliability.

 That said, it's not set in stone. It sounds like a reasonable choice to me, but I'm happy to discuss of the problematic use cases that arise.

Now for your problem of directory structure convention: I'm not sure how does it cause a problem? Unless you have 2 directories with the same name and one lower case, the other uppercase, it doesn't change anything on that side right? (also you could use `/` after that PR in the names to match the directory structure btw)

>We also have this case where its nice to list all the names in the fixtures that are canonical and generate slugs or alias based on those

That's an interesting use case but it's one where I would say you are misusing alice flexibility and should be done another way. The first problem is that it's very easy to insert special characters in the list name that way, (ex: `Alice's Cat`) which can be problematic for the reasons mentioned before. The second is that this could be easily done by having a custom faker provider where you have a whole list of those possible values instead:

```yaml
Class:
  "reference_{1..5}":
    name (unique): <slugName()>
    alias: <(someUrlFriendlyFunction($name))>
```

Now the only drawback I see is you're not able to do anymore "I know I have X values, I want a fixture for each of it" in a dynamic manner. But in that case I would argue that you could have your list of fixed fixtures as a parameter and have your fixtures matching its size:

```yaml
parameters:
  slugs: [Alice, White Rabbit, Mad Hatter, 'Alice\'s Cat']

Class:
  reference_<size(slugs)>:
    name (unique): <slugName()>
    alias: <(someUrlFriendlyFunction($name))>
```

(It doesn't work right now), but my point is I would find that more flexible and robust, and would prefer to add this new feature (which would go well with #347), than relying on your current way which looks more hacky and much more problematic if you were to get the references of your generated fixtures for example.

That said:

1. We don't plan to remove it in v2 at least, only deprecating it. So if that's something working well for you now and changing it would make things hard, there is no need to upgrade right away.
2. V3 still have a few months before being out, you got time.

matt-usurp:
Your points are very valid around testing and I would understand the reasoning if you were moving to the `symfony/config` component but currently we are parsing custom `yaml` with custom parsers. So totally understand the want for "standardising" the references I just feel like removing the case of characters might be a little much. Removing special characters I totally agree with. 

Yeah, the wonders of opening code up to people means you will get a lot of "creative use cases". Maybe this would be a feature but if it was possible that an objects content was generated before its reference then we could use variable names in the reference too. Essentially we need a way of targeting these specific instances as other fixtures rely heavily on targeting the correct data. An example being maybe a list of products and creating fixed amount of orders for these products would require us knowing the `@reference`. Our use case is that the fixture data isn't truly production data but we as close as possible. Working with numbers `{1..9}` is not except able for large datasets.

But like I raised in a previous ticket (on phone so cannot find it) the use of `current()` with the `identity()` format would solve most of my issues that I had with this change. The directory structure thing can be easily replaced with underscore and period format (that we use for service definitions anyway).

theofidry:
Ah my bad, I thought that if you had two references `userbob` and `UserBob`, `@user*` would refer to both of them but after a check looks like it is case sensitive for that one. As that currently works I don't mind keeping it then.

doc/relations-handling.md
theofidry: TBC: allow case

src/Nelmio/Alice/Fixtures/Fixture.php
```diff
@@ -373,4 +373,37 @@ protected function getConstructorComponents()
 
         return ['method' => '__construct', 'args' => $constructorValue];
     }
+    
+    private function checkName($name)
+    {
+        $lowerCaseName = strtolower($name);
+        if ($lowerCaseName !== $name) {
```
theofidry: remove that check

tests/Nelmio/Alice/Fixtures/Builder/Reference.php
theofidry: @tshelburne will you have time to review this one in the coming weeks?

theofidry:
ping @tshelburne

tshelburne:
@theofidry Can you ping me post-resolving merge conflicts? Since I haven't been close to the code as it changes, I want to make sure the picture is clear.

src/Nelmio/Alice/Fixtures/Builder/Methods/ListName.php
```diff
@@ -25,21 +25,57 @@ class ListName implements MethodInterface
...
+            return true;
+        }
+
+        return true;
     }
 
     /**
      * {@inheritDoc}
+     *
+     * @throws \InvalidArgumentException
```
theofidry: TBR

theofidry:
@tshelburne rebased. I forgot it was such a big PR... better not to let it staging too long.

theofidry:
@matt-usurp I removed the case sensitiveness check.

tests/Nelmio/Alice/Fixtures/Builder/Reference.php
tshelburne: @theofidry I like the decisions you've made and the tests you're running - I'm personally not in favor of using engineered solutions in tests (factories, etc.) since it makes things a bit more opaque, but that's just my personal taste. My preference is to have tests be stupidly obvious, because from my perspective, tests are the best documentation there is for a library.

Other than that, I like it all.

src/Nelmio/Alice/Fixtures/Builder/Methods/ListName.php
```diff
@@ -25,21 +25,50 @@ class ListName implements MethodInterface
      */
     public function canBuild($name)
     {
-        return 1 === preg_match('#\{([^,]+(\s*,\s*[^,]+)*)\}#', $name, $this->matches);
+        if (1 !== preg_match('/\{(?<content>.*)\}/', $name, $this->matches)) {
```
tshelburne: @theofidry There's no reason to use `$this->matches` here - that was being used when we were testing in the affirmative to avoid preg matching twice. Not my best code ;)

tshelburne:
@theofidry In general, this all looks good - I'm not a fan of the "magic" with data providers and all that for testing, but I get it. I'm fine either way here.

tests/Nelmio/Alice/Fixtures/Builder/Reference.php
theofidry: >I'm personally not in favor of using engineered solutions in tests (factories, etc.) since it makes things a bit more opaque

I feel the same, duplication is not bad in tests. However in this case doing it feels... overwhelming. The idea was to get a complete list of references somewhere which can be used and re-used. Sometimes the whole thing (for integration tests), sometimes just some bits (for some unit tests). The idea is that unlike the integration and unit tests, they will be portable to v3 as well.

matt-usurp:
You said you removed the case sensitive check, what does this actually mean?

Also I would say you need to work on your commit messages as looking at the commits tells me nothing about what you have done recently. Like, I can't even see a commit that might be referencing that change.

theofidry:
>You said you removed the case sensitive check, what does this actually mean?

There's been no change on that side in the end.

>Also I would say you need to work on your commit messages as looking at the commits tells me nothing about what you have done recently. Like, I can't even see a commit that might be referencing that change.

You have the details of what changed in the changelog which gives you the reference to the PR and you can always get the PR from commit.

PR #356, opened 2016-05-15
Deprecate usage of context in parsers

### LINKED ISSUE #342: Deprecate parser context for v3
After working on the parsers (#340), I noticed that a context, accessible in the parsed files, can be set at the parsers level.

For example:

```php
use Nelmio\Alice\Fixtures\Parser\Methods\Php as PhpParser;

$context = ['foo' => 'bar'];
$parser = new PhpParser($context);
```

```php
// parsed_file.php
// has access to $context
return [
  'Dummy' => [
    'dummy0' => [
      'id' => 0,
      'name' => $context['foo'],
    ]
  ]
]
```

In YAML it is a bit more tricky as you cannot use raw PHP. Instead you have to convert your yaml file into a PHP one first:

```php
use Nelmio\Alice\Fixtures\Parser\Methods\Yaml as YamlParser;

$context = ['foo' => 'bar'];
$parser = new YamlParser($context);
```

```yaml
# parsed_file.yml.php
# has access to $context
Dummy:
    dummy0:
      id: 0
      name: <?php echo $context['foo']; ?> # has to be followed by a blank line... ugly workaround

# ...
```

I find this feature questionable and it requires some workaround to make it possible. Removing it would be much simpler. The only useful usage I found was in the code base to inject the loader itself to inject the parameters. Taking a look at [this diagram](https://github.com/nelmio/alice/issues/265#issuecomment-217417501), this could be handled in the `Builder` instead.

@tshelburne WDYT?

theofidry:
Closes #342

### DIFF
--- a/src/Nelmio/Alice/Fixtures/Builder/Methods/RangeName.php
+++ b/src/Nelmio/Alice/Fixtures/Builder/Methods/RangeName.php
@@ -25,7 +25,7 @@ class RangeName implements MethodInterface
         if (1 === preg_match('/\{([0-9]+)(\.{3,})([0-9]+)\}/i', $name, $this->matches)) {
             @trigger_error(
                 'Ranged name should follow the pattern "name{X..Y}". Using "name{X...Y} or with more dots instead is '
-                .'deprecated sinece 2.2.0 and will be removed in 3.0. Please mind the change of behavior: "user{0..10}"'
+                .'deprecated since 2.2.0 and will be removed in 3.0. Please mind the change of behavior: "user{0..10}"'
                 .'is creating 11 users whereas "user{0...10}" is creating 10',
                 E_USER_DEPRECATED
             );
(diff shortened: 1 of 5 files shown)


src/Nelmio/Alice/Fixtures/Builder/Methods/RangeName.php
```diff
@@ -25,8 +25,8 @@ public function canBuild($name)
         if (1 === preg_match('/\{([0-9]+)(\.{3,})([0-9]+)\}/i', $name, $this->matches)) {
             @trigger_error(
                 'Ranged name should follow the pattern "name{X..Y}". Using "name{X...Y} or with more dots instead is '
-                .'now deprecated and will be removed in 3.0. Please mind the change of behavior: "user{0..10}" is '
-                .'creating 11 users whereas "user{0...10}" is creating 10',
+                .'deprecated sinece 2.2.0 and will be removed in 3.0. Please mind the change of behavior: "user{0..10}"'
```
reinfi: there's a little typo in your error text
sinece => since

theofidry:
Good catch ;)

Issue #291, opened 2015-11-20
Unique option not working with templates


Soullivaneuh:
The issue is quite simply to understand but was hard to found.

Basically, if you try to set unique option on a template:

```yml
AppBundle\Entity\Domain:
    domain (template):
        name (unique): <domainName()>
        status: 1
        server: @server_*
        ftpLogin: <userName()>
        mysqlLogin: <userName()>
    domain_special_chars (extends domain):
        name: sßnté-çava.æfr
    domain_{0..500} (extends domain): ~
```

This is not working, producing a doctrine constraint exception.

But, if you move back the `name` property outside the template:

```yml
AppBundle\Entity\Domain:
    domain (template):
        status: 1
        server: @server_*
        ftpLogin: <userName()>
        mysqlLogin: <userName()>
    domain_special_chars (extends domain):
        name: sßnté-çava.æfr
    domain_{0..500} (extends domain):
        name (unique): <domainName()>
```

This works properly.

It seems unique option is simply ignored with templates.

theofidry:
Related to #276 as well I guess.

theofidry:
Closed via #359

theofidry:
@Soullivaneuh could you please check if this has been solved by #359?

PR #359, opened 2016-05-16
Unique option not working with templates (issue #291)

reinfi:
PR fixes https://github.com/nelmio/alice/issues/291

### DIFF
--- a/src/Nelmio/Alice/Fixtures/Fixture.php
+++ b/src/Nelmio/Alice/Fixtures/Fixture.php
@@ -107,7 +107,7 @@ class Fixture
         }
 
         foreach ($template->properties as $property) {
-            if (!isset($this->spec[$property->getName()])) {
+            if (!$this->hasProperty($property->getName())) {
                 $this->addProperty($property->getRawName(), $property->getValue());
             }
         }
@@ -159,6 +159,21 @@ class Fixture
     public function getName()
     {
         return $this->name;
+    }
+
+    /**
+     * @param string $name
+     * @return bool
+     */
+    public function hasProperty($name)
+    {
+        foreach ($this->properties as $property) {
+            if ($property->getName() === $name) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


tests/Nelmio/Alice/support/fixtures/unique_with_template.yml
```diff
@@ -0,0 +1,6 @@
+\Nelmio\Alice\support\models\User:
```
theofidry: I think we need a more complete test for that one, e.g. add another non unique field to the template. It would be also good to have another example with two level deep, i.e.

- template1: 2 fields with unique, 1 without
- template2: extends template 1, override 1 unique field to make it not unique, add another unique field
- 3: extends template 1

Don't hesitate to create dedicated entities for it if needed

theofidry:
Again thanks a lot for looking into it! It's definitely an issue that I don't think I would have had the time to look before a while!

reinfi:
I will setup more tests and will change the implementation also so that it will be more flexible for other extensions.

theofidry:
>so that it will be more flexible for other extensions.

Don't over do it though, v3 is coming out ;)

reinfi:
good advice for the tests because I found another bug within template inheritance.

theofidry:
@reinfi what is the status of the PR? Is there any work left or is it ready for review?

reinfi:
it is rady for review

src/Nelmio/Alice/Fixtures/Fixture.php
```diff
@@ -127,11 +127,13 @@ function ($flag) {
             }
         );
 
-        return array_map(
-            function ($extension) {
-                return str_replace('extends ', '', $extension);
-            },
-            $extensions
+        return array_reverse(
```
theofidry: why the `array_reverse`? This will mess up the extending order which can be very problematic

reinfi: I removed it, I had no clue why I added it.

theofidry:
@reinfi looks like tests are failing. I guess that's what the `array_reverse` was for, but I don't think this would be the right solution as it would be a BC break.

theofidry:
You might want to rebase as well, no big change though :)

theofidry:
Closed via d1e88de. Thanks very much for this @reinfi :)

Issue #265, opened 2015-09-14
[v3] Roadmap


theofidry:
#**EDIT: update this part of the message to avoid to have to crawl the whole thread to get a review of v3.**

Main changes with v3:

- Drop the persistence layer
- Make the library more resilient
- Make the library more extendable
- Make the library more performant
- Add a few new features & bugfixes

Global overview of the library:

<img width="367" alt="screen shot 2016-06-04 at 00 08 25" src="https://cloud.githubusercontent.com/assets/5175937/15795337/8e60b0c0-29e8-11e6-9c83-10b938c8dec6.png">

Main interfaces (#381):

```php
LoaderInterface::load(string $file, array $parameters = [], array $objects = []): ObjectSet;

FixtureBuilderInterface::build(string $file, array $parameters = [], array $objects = []): FixtureSet;
GeneratorInterface::generate(FixtureSet $fixtureSet): ObjectSet;
```

- [x] `Loader` (#384)

`FixtureBuilder` secondary interfaces (#382):

- [x] `Parser`(#361, #385)
- `Denormalizer`
  - [x] `Fixture` (#394)
  - [x] `Parameter` (#386)

`Generator` secondary interfaces (#383):

- [x] Generator (#395)
- Resolver
  - [x] Main resolver (#404)
  - [x] Parameters (#366, #396)
  - [x] Fixtures (#397)
- [x] Instantiator (#440)
- [x] Specs Resolver (#449)
- [x] Populator (#442, #450)
- [ ] Configurator

–––

(original message)

What's your plans @tshelburne? I would be more than happy to work on it :)

tshelburne:
Haven't had plans right now, since 2.x is still relatively fresh. I think it would take, at minimum, a discussion of what we want to accomplish and how things are failing right now. 

From my perspective, the biggest struggle we have right now is that the "lifecycle" of moving from files -> fixtures -> instances is not particularly flexible, and all the state getting passed around and updated all over the place makes it hard to reason about making changes. I think a more well-defined interface for that flow would let us do things like the relationship hydration step more easily.

I'm going to be mostly just reviewing PRs for a while - I have some client work that's pressing right now. The best start right now would probably be to start drafting a proposed changeset and some ideas behind desired features. It would be nice to just move forward with some of the issues in a 3.x version, rather than trying to forcibly patch them into 2.x. I don't think we'll need to have *too* many BC problems for a 3.x release anyway, but that's yet to be seen.

theofidry:
[...]

PR #361, opened 2016-05-17
[v3] Rework parsers

theofidry:
As discussed in #265, here's the new parser system:

screen shot 2016-05-17 at 23 45 36

Proposed changes:

- Parsers now only parse files, they no longer hold state (like parameters) except to cache the parsing result for the Cache parser
- Parsing is no longer contextual (the loader is no longer injected in the parsed files c.f. #342) which is what makes them now cacheable
- Parsers are now defined by a `ParserInterface` interface
- `Php` and `Yaml` methods have been dropped to be real parsers instead (which can be reused as standalone, implements the `ParserInterface)`
- `Parser` has been replaced by `ParserRegistry`. The principle is the same (using the Reponsibility Chain pattern), with the difference that parsers are chainable only if they implement the new `ChainableParserInterface`
- There is a new key parser `RuntimeCacheParser`: it is the parser responsible for caching the results and proceed the includes (it is done their to cache the result of included files)

Tests should be rather clear and are also readable with testdox:

screen shot 2016-05-17 at 23 58 27

You can see the result on Travis and the code coverage is of 100% (for the parser).

/cc @Soullivaneuh @matt-usurp @reinfi


src/Nelmio/Alice/Exception/Parser/RuntimeException.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\Alice\Exception\Parser;
+
+use Nelmio\Alice\Throwable\ParseThrowable;
+
+class RuntimeException extends \RuntimeException implements ParseThrowable
```
theofidry: To be removed

src/Nelmio/Alice/Parser/ParserRegistry.php
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * @throws \InvalidArgumentException When invalid parser is passed.
+     */
+    public function __construct(array $parsers)
+    {
+        foreach ($parsers as $parser) {
+            if (false === $parser instanceof ChainableParserInterface) {
+                throw new \InvalidArgumentException(
+                    sprintf(
+                        'Expected parsers to be "%s" objects. Parser "%s" is not.',
```
theofidry: ~~`Parser "%s" is not.`~~ `Got "%s" instead.`

src/Nelmio/Alice/Parser/ParserRegistry.php
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public function parse(string $file): array
+    {
+        foreach ($this->parsers as $parser) {
+            if ($parser->canParse($file)) {
+                return $parser->parse($file);
+            }
+        }
+
+        throw new RuntimeException(
```
theofidry: throw a `ParserNotFoundException` (which could extend `\RuntimeException`)

src/Nelmio/Alice/Parser/RuntimeCacheParser.php
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * @param string $file File loaded
+     * @param array  $data Parse result of the loaded file
+     *
+     * @throws InvalidArgumentException
+     * 
+     * @return array
+     */
+    private function processInclude(string $file, array $data): array
```
theofidry: Should be extracted in another parser or class dedicated to process the includes. The reason being is that the way the filepath is being resolved should be customizable; cf. https://github.com/hautelook/AliceBundle/issues/203

PR #365, opened 2016-05-21
Adding a link to a tutorial about using raw Alice with Symfony

weaverryan:
Of course, this isn't a library and it's under that section :), but this is where I would look if I was searching for "how do I use Alice in Symfony"

Cheers!


theofidry:
@weaverryan what about adding it under the ToC like for [AliceBundle#documentation](https://github.com/hautelook/AliceBundle#documentation)? Would be less confusing and more visible IMO

weaverryan:
You beat me to it :). Thanks!

PR #368, opened 2016-05-23
Fix typo in processors.md

kaoru:
Noticed this typo while reading the docs. Could be that the typo is duplicated elsewhere in the docs, I didn't have a chance to do a proper check through, but thought reporting this was better than not doing so :-)


theofidry:
Thanks @kaoru, after a quick search looks like it was the only occurrence :)

PR #401, opened 2016-06-17
Add link to trappar/AliceGeneratorBundle

trappar:
Just published version 0.1.0 of this library which offers the ability to recursively convert existing Doctrine entities into Alice fixtures. Seems like a resource which fits right in with the other third part libraries 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ For more information, refer to [the documentation](#table-of-contents).
   * [hautelook/AliceBundle](https://github.com/hautelook/AliceBundle)
   * [h4cc/AliceFixturesBundle](https://github.com/h4cc/AliceFixturesBundle)
   * [knplabs/rad-fixtures-load](https://github.com/KnpLabs/rad-fixtures-load)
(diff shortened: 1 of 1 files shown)


theofidry:
As it's tightly coupled to AliceBundle it should stay in AliceBundle IMO, not further.

trappar:
The bundle itself doesn't require any component from AliceBundle to run, it only uses AliceBundle for testing. I haven't tested it with other libraries for loading the fixtures yet, but since all it does is generate normal Alice fixture yaml files, it should theoretically work with anything that loads Alice fixtures. Maybe I'll open a new pull request after I've done further testing.

theofidry:
>The bundle itself doesn't require any component from AliceBundle to run

Then it shouldn't be a bundle :) You could rename it and make it a true library, and in it have a bridge for Symfony, e.g. like done in api-platform/core.

If it's published a PHP package and not a Symfony bundle then I see no reason not to include it in alice, but you should deliver that promise first (i.e. change the package name and test it without AliceBundle dependency)

trappar:
I'll look into it. Thanks 😄

PR #407, opened 2016-06-22
Fixed deprecation warnings

### LINKED ISSUE #389: Deprecation warning for singularify in Symfony 3.1
Looks like a simple fix.

`StringUtil::singularify() is deprecated since version 3.1 and will be removed in 4.0. Use Symfony\Component\Inflector\Inflector::singularize`

pkruithof:
Fixes #389, and also warnings from phpunit.

I also added a psr/log dependency, as it's being used in the Loader class.

### DIFF
--- a/src/Nelmio/Alice/Instances/Populator/Methods/ArrayAdd.php
+++ b/src/Nelmio/Alice/Instances/Populator/Methods/ArrayAdd.php
@@ -11,9 +11,6 @@
 
 namespace Nelmio\Alice\Instances\Populator\Methods;
 
-use Symfony\Component\Form\Util\FormUtil;
-use Symfony\Component\PropertyAccess\StringUtil;
-
 use Nelmio\Alice\Fixtures\Fixture;
 use Nelmio\Alice\Util\TypeHintChecker;
 
@@ -24,9 +21,19 @@ class ArrayAdd implements MethodInterface
      */
     protected $typeHintChecker;
 
+    /**
+     * @var callable
+     */
+    private $singularizer;
+
+    /**
+     * @param TypeHintChecker $typeHintChecker
+     */
     public function __construct(TypeHintChecker $typeHintChecker)
     {
         $this->typeHintChecker = $typeHintChecker;
+
+        $this->detectSingularizeMethod();
     }
 
     /**
@@ -52,9 +59,10 @@ class ArrayAdd implements MethodInterface
     /**
      * finds the method used to append values to the named property
      *
-     * @param  mixed  $object
-     * @param  string $property
-     * @return string Method name or null if adder method not detected
+     * @param mixed  $object
+     * @param string $property
+     *
+     * @return string|null Method name or null if adder method not detected
      */
     private function findAdderMethod($object, $property)
     {
@@ -62,17 +70,9 @@ class ArrayAdd implements MethodInterface
             return $method;
         }
 
-        if (class_exists('Symfony\Component\PropertyAccess\StringUtil') && method_exists('Symfony\Component\PropertyAccess\StringUtil', 'singularify')) {
-            foreach ((array) StringUtil::singularify($property) as $singularForm) {
-                if (method_exists($object, $method = 'add'.$singularForm)) {
-                    return $method;
-                }
-            }
-        } elseif (class_exists('Symfony\Component\Form\Util\FormUtil') && method_exists('Symfony\Component\Form\Util\FormUtil', 'singularify')) {
-            foreach ((array) FormUtil::singularify($property) as $singularForm) {
-                if (method_exists($object, $method = 'add'.$singularForm)) {
-                    return $method;
-                }
+        foreach ($this->singularize($property) as $singularForm) {
+            if (method_exists($object, $adder = 'add'.$singularForm)) {
+                return $adder;
             }
         }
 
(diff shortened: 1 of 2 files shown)


theofidry:
Thanks for the work @pkruithof!

> I also added a psr/log dependency, as it's being used in the Loader class.

Could you remove that part? Logger is an optional dependency and the library works fine without it :)

pkruithof:
Done.

src/Nelmio/Alice/Instances/Populator/Methods/ArrayAdd.php
```diff
@@ -84,4 +85,37 @@ private function findAdderMethod($object, $property)
             return $method;
         }
     }
+
+    /**
+     * @param string $plural
+     *
+     * @return string
```
theofidry: can return null if no result

src/Nelmio/Alice/Instances/Populator/Methods/ArrayAdd.php
```diff
@@ -62,16 +69,10 @@ private function findAdderMethod($object, $property)
...
-                if (method_exists($object, $method = 'add'.$singularForm)) {
-                    return $method;
-                }
-            }
-        } elseif (class_exists('Symfony\Component\Form\Util\FormUtil') && method_exists('Symfony\Component\Form\Util\FormUtil', 'singularify')) {
-            foreach ((array) FormUtil::singularify($property) as $singularForm) {
-                if (method_exists($object, $method = 'add'.$singularForm)) {
-                    return $method;
+        if ($singulars = $this->singularize($property)) {
+            foreach ((array) $singulars as $singularForm) {
```
theofidry: the `(array)` is to take care of the case where `$this->singularize($property)` returns `null`? If so I would rather just make sure we return an array in `::singularize()`

src/Nelmio/Alice/Instances/Populator/Methods/ArrayAdd.php
```diff
@@ -62,16 +69,10 @@ private function findAdderMethod($object, $property)
             return $method;
```
theofidry: Could you also fix the return type in the phpdoc to `string|null`?

src/Nelmio/Alice/Instances/Populator/Methods/ArrayAdd.php
```diff
@@ -84,4 +85,37 @@ private function findAdderMethod($object, $property)
...
+    /**
+     * @param string $plural
+     *
+     * @return string
+     */
+    private function singularize($plural)
+    {
+        if (null === $this->singularizeClass || null === $this->singularizeMethod) {
+            // default to false so we only try this once
+            $this->singularizeClass = false;
```
theofidry: why does `singularizeClass` and `singularizeMethod` should be class properties? Couldn't they be regular variables?

pkruithof: Because this way it only has to be done once. Reflection is a costly operation, in terms of performance. I tend to always minimise the amount of reflection being done. This is assuming that an instance of this class is being used for multiple calls. If not then a static class property would be better, I think.

pkruithof:
I'll apply your other feedback tomorrow when I have more time.

src/Nelmio/Alice/Instances/Populator/Methods/ArrayAdd.php
theofidry: Ah fair I didn't think of that use case. That said, we are using only 1 instance of `ArrayAdd`, so if left as it is you could end up using the reflection class of another class than the one checked.

pkruithof: Not sure what you mean by that last part. It's better to use the same and preferred inflection method that's available, right?

theofidry: As you have only 1 instance of `ArrayAdd` used for all objects, if you leave two objects `Dummy` and `AnotherDummy`, when calling `findAdderMethod()` on a `Dummy` object, you will have `ArrayAdd#singularizeClass === 'Dummy'`. This is fine if the next object called is a `Dummy` object, but when the next object will be `AnotherDummy`, `findAdderMethod()` will return a false result.

pkruithof: What's being cached is the class/method to singularize a string, not the adder for an object. As the singularize method is called for every ArrayAdd operation this can really add up, depending on your fixtures.

theofidry: Oh I see my bad. Erm in that case this could done in the constructor? If done in `findAdderMethod()` it has the benefit to be done only if required but you have an if check on each usage.

src/Nelmio/Alice/Instances/Populator/Methods/ArrayAdd.php
```diff
@@ -83,5 +83,42 @@ private function findAdderMethod($object, $property)
...
+            'Symfony\Component\PropertyAccess\StringUtil' => 'singularify',
+            'Symfony\Component\Form\Util\FormUtil' => 'singularify',
+        ];
+
+        foreach ($classes as $class => $method) {
+            if (class_exists($class) && method_exists($class, $method)) {
+                // cache the singularize method
+                $this->singularizer = [$class, $method];
+
+                break;
```
theofidry: in this case a `return` would be more clear IMO

src/Nelmio/Alice/Instances/Populator/Methods/ArrayAdd.php
```diff
@@ -83,5 +83,42 @@ private function findAdderMethod($object, $property)
...
+    }
+
+    /**
+     * @param string $plural
+     *
+     * @return string[]
+     */
+    private function singularize($plural)
+    {
+        if ($this->singularizer) {
```
theofidry:
```
return (null !== $this->singularizer)
  ? (array) call_user_func($this->singularizer, $plural)
  : []
;
```

src/Nelmio/Alice/Instances/Populator/Methods/ArrayAdd.php
```diff
@@ -24,9 +21,19 @@ class ArrayAdd implements MethodInterface
      */
     protected $typeHintChecker;
 
+    /**
+     * @var callable
```
theofidry: `@var <string, string> Class method pair`

theofidry:
Alright, a last change and it's ready to go :)

PR #416, opened 2016-07-03
Fix identity example

IceMaD:
Just lost 30 minutes because I copy-pasted ...

With single quote, i had
`Unexpected characters near "current')`


theofidry:
Thanks @IceMaD, I'll have to double check the whole doc for such mistakes I guess.

Issue #421, opened 2016-07-11
Version 2.2 fails loading fixtures


overgapo:
After recent update to 2.2 version i've got this:

>
[Hautelook\AliceBundle\Alice\DataFixtures\LoadingLimitException]                            
  Loading files limit of 50 reached. Could not load the following files 
/code/src/AppBundle/DataFixtures/ORM/changelogs.yml:                                        
   - Entity with ID @user-1 and class Mailer\AuthBundle\Entity\User not found

And the same for every fixtures file, where references are present.

Version 2.1.4 works fine. Has anyone faced this problem too?

user-1:
My user id, but I don't know anything about it!

theofidry:
@overgapo will look into it tonight

theofidry:
I see that HautelookAliceBundle is indeed failing: https://travis-ci.org/hautelook/AliceBundle/jobs/143354359. I have an idea on what may cause this I may be able to fix that soon :)

Soullivaneuh:
Can confirm the issue with hautelook/alice-bundle too:

```bash
$ ./docker-console make app_fixtures
./app/console --no-interaction --env=dev hautelook_alice:doctrine:fixtures:load --purge-with-truncate --bundle=AppBundle --manager=default
  > fixtures found:
      - /code/src/AppBundle/DataFixtures/ORM/incident.yml
      - /code/src/AppBundle/DataFixtures/ORM/changelog.yml
      - /code/src/AppBundle/DataFixtures/ORM/ticket.yml
      - /code/src/AppBundle/DataFixtures/ORM/payment_method.yml
      - /code/src/AppBundle/DataFixtures/ORM/puppet.yml
      - /code/src/AppBundle/DataFixtures/ORM/valid_tld.yml
      - /code/src/AppBundle/DataFixtures/ORM/package.yml
      - /code/src/AppBundle/DataFixtures/ORM/credit_card.yml
      - /code/src/AppBundle/DataFixtures/ORM/operation.yml
      - /code/src/AppBundle/DataFixtures/ORM/cron.yml
      - /code/src/AppBundle/DataFixtures/ORM/news.yml
      - /code/src/AppBundle/DataFixtures/ORM/email.yml
      - /code/src/AppBundle/DataFixtures/ORM/group.yml
      - /code/src/AppBundle/DataFixtures/ORM/server.yml
      - /code/src/AppBundle/DataFixtures/ORM/user_browser.yml
[...]

PR #431, opened 2016-07-15
Fix support for hyphens

theofidry:
https://github.com/nelmio/alice/pull/355 introduced a BC which is removing support for the usage of hyphens (`-`) in fixture names. This commit revert that change and add tests for it.

Closes https://github.com/nelmio/alice/issues/421

/cc @Soullivaneuh


src/Nelmio/Alice/Instances/Populator/Populator.php
```diff
@@ -89,7 +89,8 @@ public function populate(Fixture $fixture)
 
             $value = $property->requiresUnique() ?
                 $this->generateUnique($fixture, $property) :
-                $this->processor->process($property, $fixture->getSetProperties(), $fixture->getValueForCurrent());
+                $this->processor->process($property, $fixture->getSetProperties(), $fixture->getValueForCurrent())
+            ;
```
Soullivaneuh: I'm not sure this was wanted.

Soullivaneuh:
Works like a charm, thank! :heart: 

What about deprecation notices?

theofidry:
@Soullivaneuh they are still deprecated, unless it proves to be very easy, I don't plan to support it in v3, at least for the first iteration.

Soullivaneuh:
Yes, I'm talking about deprecation notices! 😉

Sullivan SENECHAL

theofidry:
Not sure what you mean, hyphens are properly deprecated: https://github.com/nelmio/alice/blob/2.x/src/Nelmio/Alice/Fixtures/Fixture.php#L391

PR #526, opened 2016-09-03
[Doc] Update documentation for contributing

theofidry:
@tshelburne would be possible to get a review for this one?

My plans now are to take care of some TODOs before releasing the first beta. Then I'll have to port some things in v2 and keep adding deprecation notices.

In parallel I would like to update the doc but I'm less sure about that: should I change some sections saying it's v3 specific knowing v3 isn't out yet or should I just add a warning at the beginning of the readme saying that the doc will start to be changed for v3 and give a link to the v2 doc?


tshelburne:
I'm assuming you mean the docs for master - I would go ahead and mark it with "This documentation is being updated for the v3 release, refer [here] for current stable docs". 

As for this bit, go ahead and merge, and I'll submit a PR to clean up some of the language. Just ping me afterwards and I'll jump on it.

theofidry:
You're assuming right :) I won't go back on this one before Monday so if
you want to change a few things before feel free to. I don't expect the
release of the beta before the middle or the end of next week.

PR #548, opened 2016-09-13
added Doctrine inflector as possibility for sigularizing

hboomsma:
Doctrine also has an inflector which is able singularize words. Since most users will already depend
on doctrine it is opportunistic to also support this inflector. I added it at the bottom of the array to give precedence to the already existing ones.


theofidry:
Thanks @hboomsma, is there any way to port that to 3.x as well?

hboomsma:
@theofidry, I looked for any reference to `Inflector` and `singularize` in the master branch, but I could not find any. So I am not quite sure where to "fix" this for 3.x. If you can give me a hint, I am more then willing to help out.

PR #599, opened 2016-11-03
Add support for named parameters in constructor and functions

### LINKED ISSUE #594: Support constructor named properties
For:

``` yaml
User:
    user0:
        __construct:
            - bob
            - Bob
            - '1980-10-10'
            - bob@example.org
            - 42
```

It would be better if we could do:

``` yaml
User:
    user0:
        __construct:
            username: bob
            fullname: Bob
            dateOfBirth:  '1980-10-10'
            email: bob@example.org
            favoriteNumber: 42
```

Besides being easier to understand (IMO), it also gives the possibility to re-order the properties. That said this does rely on the name of PHP variable, i.e. the constructor must look like this:

``` php
class User
{
  public function __construct($username, $fullname, $dateOfBirth, $email, $favoriteNumber)
  {
    //...
  }
}
```

And changing a variable name _can_ break the fixture code.

@tshelburne WDYT?

theofidry:
Closes #594.

- [ ] Add tests for the support in functions
- [ ] Add tests for support with the usage of flags
- [ ] Finish the PR (add missing tests + cleanup)

@tshelburne could you please double check the integration tests (in `LoaderIntegrationTest` to see if that makes sense to your or if I'm missing something? Cheers

### DIFF
--- a/fixtures/Entity/Instantiator/AmbiguousDummyWithConstructorWithArrayParameter.php
+++ b/fixtures/Entity/Instantiator/AmbiguousDummyWithConstructorWithArrayParameter.php
@@ -1,0 +1,42 @@
+<?php
+
+/*
+ * This file is part of the Alice package.
+ *
+ * (c) Nelmio <hello@nelm.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Nelmio\Alice\Entity\Instantiator;
+
+class AmbiguousDummyWithConstructorWithArrayParameter
+{
+    /**
+     * @var bool
+     */
+    private $constructor = false;
+
+    /**
+     * @var array
+     */
+    private $value;
+
+    public function __construct(array $value)
+    {
+        $this->value = $value;
+        $this->constructor = true;
+    }
+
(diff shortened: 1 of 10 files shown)


tshelburne:
@theofidry The test itself looks good to me after a quick review. Thorough job.

theofidry:
Ok major disappointment, this cannot be done unless in a very limited context:

- In the case where the factory is a service, the service is resolved at generation time while we need info (reflection) on the factory method during the denormalization
- In the case of a function, the point above is even more true
- The two points above are still true (although becomes doable) if the factory/method is a static call on another function

I'm also not found of the idea of having it working in 1 case and not elsewhere, that would only be confusing IMO.

julienfalque:
I was about to open a feature request for this but Github helped me find #594 :)

Is [the reason this was aborted](https://github.com/nelmio/alice/pull/599#issuecomment-258318056) still true? I don't know the internals of the library, how important is it for named arguments to be resolved at denormalization time? Can't this be done at instanciation time (when the factory is supposed to be resolved)?

theofidry:
I think that maybe could be done by passing the names to the denormalized object and at instantiation doing extra work to handle those names instead of a simple array of arguments

julienfalque:
Do you mind reopening #594? I might give it a try if I find time.

PR #600, opened 2016-11-04
Mark the expression language as internal

theofidry:
The Expression Language is still relatively new and in-house. It may be possible that another library like https://github.com/hoaproject/Compiler may help a lot and replace a huge chunk of it. As such, I don't think I want to make this part of the library public and be potentially able to change it even after a stable release.

That said, this part of the API leaks to the [ValueResolvers](https://github.com/nelmio/alice/tree/master/src/Generator/Resolver/Value/Chainable), which makes them rather unstable. So I am not sure if this is actually doable.

@tshelburne thoughts?


tshelburne:
The idea makes sense to me - it's always good to clarify intentions. I would say just create a ticket with RFC to look at alternatives, like the compiler above, and mention the issue with value resolvers there. That way the problem is memorialized for future work, but we can keep things as clear as possible in the present.

PR #610, opened 2016-11-07
Deprecate calling methods

### LINKED ISSUE #372: Deprecate usage of method calls and custom setters
As of #313, [calling methods](https://github.com/nelmio/alice/blob/3bf8e6f9cabb4f7d06d0c0d8f715ceb4b41d19c9/doc/complete-reference.md#calling-methods) and [Customer setters](https://github.com/nelmio/alice/blob/3bf8e6f9cabb4f7d06d0c0d8f715ceb4b41d19c9/doc/complete-reference.md#custom-setter) should be deprecated.

theofidry:
Closes #372.

### DIFF
--- a/src/Nelmio/Alice/Instances/Populator/Methods/ArrayDirect.php
+++ b/src/Nelmio/Alice/Instances/Populator/Methods/ArrayDirect.php
@@ -39,6 +39,12 @@ class ArrayDirect implements MethodInterface
      */
     public function set(Fixture $fixture, $object, $property, $value)
     {
+        @trigger_error(
+            'Using a method call is deprecated since 2.3.0. This feature is will be removed in 3.0.0 in favour of'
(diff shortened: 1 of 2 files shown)


mcfedr:
So this is deprecated in 2.3, meaning phpunit will fail - but its not possible to use the new syntax in 2.3 so there is no way of avoiding this deprecation...

theofidry:
>So this is deprecated in 2.3, meaning phpunit will fail

PHPUnit will fail only if you configure it to fail with deprecation errors.

> but its not possible to use the new syntax in 2.3 so there is no way of avoiding this deprecation...

Exact, you need to upgrade to fix it. I'll gladly accept any PR to back-port that feature in 2.3 allowing a smoother upgrade, but I'm really not willing to work on it.

Issue #492, opened 2016-08-22
How to do arithmetic


thedamnedrhino:
I'm trying to do some basic arithmetic, using things like `<current() * 10>`, but I only get `0` as result. Actually `0` is what's saved in the database so that might return `null`. So how is it possible to do basic arithmetic Operations in the Fixtures?

theofidry:
`<something()`>  is the notation for calling the function `something`, so `<current() * 10>` will not work and should even throw an error IMO. I am not sure why this is not the case in 2.x but it will throw an error for sure in 3.x.

To achieve what you want to do, you need to have a custom function to do the operations. It is a bit inconvenient and might be something to enhance in the future.

theofidry:
Update on the issue, this is easily doable by using the identity function: `<(10 * 5)>`. The only limitations are that the identity function only knows about references and variables: `<(@user->favoriteNumber * $current)>` (the `<current()>` is accessible via the `$current` variable).

Issue #374, opened 2016-05-24
how to use current() in identity()


scuben:
I found the documentation about identity and its use of variables in it. I also found several issues regarding using current() in identity(). Still I did not manage to get it working.

As far as I could read in the issues it has been implemented. So... Can anyone provide an example? As a bonus you could in that case directly add a PR explaining it in the docs.

theofidry:
Hello,

`identity()` is a simple "faker" function, not provided by faker natively, but by a [Faker provider](https://github.com/nelmio/alice/blob/3bf8e6f9cabb4f7d06d0c0d8f715ceb4b41d19c9/doc/customizing-data-generation.md#customizing-data-generation) still: [IdentityProvider](https://github.com/nelmio/alice/blob/3bf8e6f9cabb4f7d06d0c0d8f715ceb4b41d19c9/src/Nelmio/Alice/Instances/Processor/Providers/IdentityProvider.php).

You have a piece of doc [here](https://github.com/nelmio/alice/blob/3bf8e6f9cabb4f7d06d0c0d8f715ceb4b41d19c9/doc/customizing-data-generation.md#default-providers), but I agree it could be better. However here's some examples:

```yaml
Nelmio\Entity\User:
  user1:
    username: 'testuser'
    fullname: '<identity($username)>'
  user2:
    username: 'test_user'
    fullname: '<identity(str_replace("_", " ", $username))>'
```

`<current()>` is also explained [here](https://github.com/nelmio/alice/blob/3bf8e6f9cabb4f7d06d0c0d8f715ceb4b41d19c9/doc/relations-handling.md#references). Same I think it could be better so any PR to help is welcomed :)

In short, `<current()`> can be used when dealing with a fixture collection such as `user_{1..10}` or `user_{alice, bob}` and returns the "current" value of the collection:

```yaml
Nelmio\Entity\User:
  user_{1..10}: # creates 10 users "user_1", ..., "user_10"
    username: '<current()>' # will be "1", ..., "10"
  user_{alice, bob}: # creates two users "user_alice" and "user_bob"
    username: '<current()>' # respectively "alice" and then "bob"
```

If my memory serves me right, it throws an exception if you try to use `<current()>` outside of that context (needs to be checked).

Best,

scuben:
Hi @theofidry 

Thanks for your response. I know both features `<identity()>` and `<current()>`. But what I still dont know about how to use both at the same time. I try to have `<current()>` subtracted with 2.

[...]

Issue #360, opened 2016-05-16
current() usage in identity()?


matt-usurp:
Is it possible (and if so its not clear from the documentation) to make use of the `<current()>` syntax within something like `<(strtolower(??))>`?

Essentially I want to be able to be able to get this:

```YAML
SomeClass:
  SomeReference{Name}:
    name: <current()>
    alias: <(strtolower($current))>
```

theofidry:
Well `current` is a function so it would be more `<(strtolower(current()))>`. That said, I don't think it works and I didn't check how reliable chain calls, i.e. `<(strtolower(anotherfunction($var)))>` behaves (although I would like it to work).

That said, in light of #355 you should avoid upper case names :)

matt-usurp:
As a quick note, I am aware that `current()` is a function and maybe I should have mentioned it. But when wrapped in the `identity()` function or `()` it is evaluated as the array function for getting the current pointer in an array. So I used `$current` as the value is kind of variable. But yeah, some interesting use case. I was going to try dig out where the code is evaluated and see if the current index or list entry is within local scope to grab. Maybe not though.

theofidry:
To be honest, it's still a dark part of the project for me. I'm getting more familiar with it as I'm adding tests, debugging it and writing v3 which makes me ports a few changes to v2 or help me to find an easy fix on v2. But I'm not at the data generation yet so I can't be of much help at the moment.

theofidry:
After checking https://github.com/nelmio/alice/issues/374, I think `<strtolower(current())>` works (couldn't check, but that would be the syntax supposed to work).

theofidry:
Indeed it looks like it's not working :(

piotrkochan:
+1 please fix this. Also it'll be nice to be able to concatenate `current()` result with string, eg:

    <randomFunction("somestring-{current()}")>

theofidry:
@piotrkochan sadly there is no plan to fix it on the short-term. Not that I don't want to, but I suspect the problem to be deeper and being very hard to fix especially without introducing any BC break. I might be wrong, but I prefer to dedicate the little time that I have on v3 for which I can guarantee that this will work rather than trying to fix v2.

Of course if someone is ready to spend time to investigate it and provide a bugfix, I'll gladly accept it, but understand that it is not my priority at the moment.

theofidry:
[...]

PR #612, opened 2016-11-11
$current in identity() #374 #360 #492

karser:
I wasn't able to use `$current` in identity() like it's described in the [docs](https://github.com/nelmio/alice/blob/master/doc/customizing-data-generation.md#identity)
```field: <($current)>```

### CODE REVIEW

**Comment by theofidry** [comment]
Thanks for the report! I'll take a better look at that later. Meanwhile, would it be possible to have a sample of your fixtures where you're using it?

**Comment by karser** [comment]
The example might be a day of week
```
    assign{1..100}:
       dayOfWeek: '<($current % 7)>'
```

theofidry:
Indeed there is no reason why it shouldn't work in that context, I'll double check there is not a test for it somewhere which would reveal to be falsy and need to be changed

theofidry:
Closing in favour of #619

Issue #627, opened 2016-11-17
Investigate caching


theofidry:
As it has been design in #265, the result generated by the `FixtureBuilder` is a comprehensible graph of fixtures which is invariant: loading X times the same set of data with the `FixtureBuidler` will result each time in the same graph of fixtures.

However, we can see here (profile generated by the scenario 2 in `profiling`:

<img width="363" alt="screen shot 2016-11-17 at 09 21 54" src="https://cloud.githubusercontent.com/assets/5175937/20383388/7685a16c-aca7-11e6-89a0-bba48b1d6041.png">

that it accounts for only ~7% of the time (it hardly gets more expansive) which was in that profile ~2.3s.

As such, I don't believe there is much to do in terms of caching on that side. However one thing that might be more interesting, is caching the whole result. Indeed if Faker seed is unchanged and not `null`, provided the user is not doing some weird random things, the result is expected to be the same.

There is however the question on how to do it: the input is faker generator see which is not necessarily injected directly in the input and an array of PHP data. And the result is a graph of objects.

theofidry:
Note that the cache needs to take into account wether or not the code source for the entities changed which makes it extremely complicated. Maybe an easier way would be to not care about that and allow the user to clear the cache manually when necessary.

PR #651, opened 2016-12-02
[Needs help] Add a cache layer

theofidry:
Attempt to tackle #627.

The implementation **will** change, things to think about:

- Should it be enabled by default?
- How to clear the cache?
- Add extensive tests, will probably require blackfire SDK
- Should it be made optional? If yes will require some tweaks to be enabled by default still

Update: I'm not really interested in this work. It is a nice bonus but I don't have enough time to spend on this. If someone needs performances and is willing to spend a bit of his time, this PoC is a very good start.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,8 +20,10 @@
     ],
     "require": {
         "php": "^7.0",
+        "doctrine/cache": "^1.6",
         "fzaninotto/faker": "^1.0",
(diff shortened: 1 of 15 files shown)


theofidry:
Closing this one. I'm letting the original issue #627 but I have no interest in continuing the PR so this is unlikely to happen unless someone is willing continue it.

PR #666, opened 2017-01-29
Implemented method calling via __calls

alekitto:
Added a `SimpleCaller` class to handle method calls via `__calls` as documented here: https://github.com/nelmio/alice/blob/master/doc/complete-reference.md#calling-methods


tests/Generator/Caller/SimpleCallerTest.php
```diff
@@ -0,0 +1,189 @@
...
+            null,
+            $fixtures = (new FixtureBag())->with(
+                $fixture = new SimpleFixture(
+                    'dummy',
+                    Dummy::class,
+                    new SpecificationBag(
+                        null,
+                        new PropertyBag(),
+                        (new MethodCallBag())
+                            ->with(new SimpleMethodCall('setTitle', [$titleValue = new FakeValue()]))
```
theofidry: I would add a second call on `setTitle` to make sure the second overrides the first

tests/Generator/Caller/SimpleCallerTest.php
```diff
@@ -0,0 +1,189 @@
...
+        $expected = new ResolvedFixtureSet(
+            new ParameterBag(['iteration' => 1]),
+            $fixtures,
+            new ObjectBag(['dummy' => $object])
+        );
+
+        $caller = new SimpleCaller($resolver);
+        $actual = $caller->doCallsOn($object, $set, $context);
+
+        $this->assertEquals($expected, $actual);
```
theofidry: an assertion on the number of calls made on the resolver is missing

src/Generator/Caller/SimpleCaller.php
```diff
@@ -0,0 +1,107 @@
...
+                try {
+                    $result = $this->resolver->resolve($value, $fixture, $fixtureSet, $scope, $context);
+                } catch (ResolutionThrowable $throwable) {
+                    throw UnresolvableValueDuringGenerationExceptionFactory::createFromResolutionThrowable($throwable);
+                }
+                list($value, $fixtureSet) = [$result->getValue(), $result->getSet()];
+            }
+        }
+
+        return $methodCall->withArguments($arguments);
```
theofidry: you should return `[$methodCall->withArguments($arguments), $fixtureSet]`. The reason `$result` contains both the a value and the set is that the set may have changed during the resolution. A simple example:

```yaml
Dummy:
  dummy2:
    __construct:
      - '@dummy1' # while resolving the value, the fixture `@dummy1` will be generated
  dummy1:
    foo: 'foo'
```

And if the `$fixtureSet` is not properly updated, when it will be the turn of `dummy1`, instead of finding the object already generated, it will generated again (which will also break the dependency between `dummy2` and `dummy1` as `dymmy2#dummy1` will no longer be `dummy1`)

theofidry:
Thanks for the PR @alekitto, that's some big work :) There is a few changes still to do, but the hardest has been made.

How was the dive in the code? I tried to make it readable but if you have some feedbacks and things that looked really weird don't hesitate to point them out!

alekitto:
I made some changes, tomorrow i'll write an integration test.
The code looks quite readable and it was pretty easy to implement and find where to modify the Loader to call the new `SimpleCaller` class.
The only thing that was unclear to me is where to find services definitions for Symfony: I didn't understand that the caller was a service and not instantiated from the Loader. Have I missed something on the doc?

theofidry:
>The only thing that was unclear to me is where to find services definitions for Symfony: I didn't understand that the caller was a service and not instantiated from the Loader. Have I missed something on the doc?

The services can be found in [src/Bridge/Symfony/Resources/config](https://github.com/nelmio/alice/tree/master/src/Bridge/Symfony/Resources/config). Basically `NativeLoader` is a loader to be able to easily use alice out of the box, without any DIC. The Symfony bridge however ensures everything is registered as services to the Symfony DIC (via the bundle `NelmioAliceBundle`) to be able to take advantage of the features the DIC provides.

It's like a traditional bundle, but shipped here instead of a dedicated library (it's easier for maintenance on my end).

src/Generator/Caller/SimpleCaller.php
alekitto: `$fixtureSet` here is passed by-ref, so it is already updated.
Do you prefer returning an array with the new fixtureSet instead of passing it by reference?

theofidry: Ah I missed that, yes please, I prefer explicit returned values over passing them by references :)

theofidry:
@alekitto tell me when you need a last review, AFAICS there's only one comment left and should be ready to ship.

src/Generator/Caller/SimpleCaller.php
```diff
@@ -0,0 +1,107 @@
...
+        GenerationContext $context
+    ): ResolvedFixtureSet {
+        if (null === $this->resolver) {
+            throw ResolverNotFoundExceptionFactory::createUnexpectedCall(__METHOD__);
+        }
+
+        $fixture = $fixtureSet->getFixtures()->get($object->getId());
+        $calls = $fixture->getSpecs()->getMethodCalls();
+
+        $scope = [
```
alekitto: I think it is needed for value resolution. Isn't it?

theofidry: Ah right it's needed for the `EvaluatedValueResolver`. Could you please add a test case for that in the integration test to not forget about it? Something like this should trigger it:

```yaml
Dummy:
  dummy1: {}
  dummy2:
    __calls:
      - setRelatedDummy: [ '<(@dummy1)>' ]
```

alekitto:
Should be all done now

theofidry:
Thanks very much @alekitto, that's a nice one :)

Issue #668, opened 2017-02-07
Call referenced object method does not resolving


aistis-:
I am trying to call a method from a referenced object

```
Entity/Random:
    random_1:
        sku: '@product_1->getSku()'
```

But I get an error in beta 3:

```
[Nelmio\Alice\Throwable\Exception\Generator\Resolver\ResolverNotFoundException]  
  No resolver found to resolve value "@product_1->getSku()". 
```

But it works `@product_1->sku` on private property.

theofidry:
@aistis- can you give the which beta version is being used as well? (there is 3 beta releases) If you're not sure, the version hash will do

aistis-:
```
"version": "v3.0.0-beta.3",
"reference": "3e53619db2590bfa72552e724cad1613a6c5c9c0"
```

theofidry:
Thanks, I'll try to have a look shortly

alekitto:
The expression is correctly parsed, but seems that a `FixtureMethodCallValue` is missing

theofidry:
I'm surprised it's missing but that's how it looks.

aistis-:
Is it something I should hook up manually?

alekitto:
Workin' on it... I will submit a PR shortly

theofidry:
@aistis- it should work on `3.x@dev` now :)

aistis-:
Thank you!

PR #669, opened 2017-02-07
Add method call reference resolver

alekitto:
See #668

### DIFF
--- a/src/Generator/Resolver/Value/Chainable/FixtureMethodCallReferenceResolver.php
+++ b/src/Generator/Resolver/Value/Chainable/FixtureMethodCallReferenceResolver.php
@@ -1,0 +1,115 @@
+<?php
+
+/*
+ * This file is part of the Alice package.
+ *
+ * (c) Nelmio <hello@nelm.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Nelmio\Alice\Generator\Resolver\Value\Chainable;
+
+use Nelmio\Alice\Definition\Value\FixtureMethodCallValue;
+use Nelmio\Alice\Definition\ValueInterface;
+use Nelmio\Alice\Throwable\Exception\Generator\Resolver\NoSuchMethodExceptionFactory;
+use Nelmio\Alice\Throwable\Exception\Generator\Resolver\NoSuchPropertyException;
+use Nelmio\Alice\Throwable\Exception\Generator\Resolver\ResolverNotFoundExceptionFactory;
+use Nelmio\Alice\Throwable\Exception\Generator\Resolver\UnresolvableValueException;
+use Nelmio\Alice\FixtureInterface;
+use Nelmio\Alice\Generator\GenerationContext;
+use Nelmio\Alice\Generator\ResolvedFixtureSet;
+use Nelmio\Alice\Generator\ResolvedValueWithFixtureSet;
+use Nelmio\Alice\Generator\Resolver\Value\ChainableValueResolverInterface;
+use Nelmio\Alice\Generator\ValueResolverAwareInterface;
+use Nelmio\Alice\Generator\ValueResolverInterface;
+use Nelmio\Alice\IsAServiceTrait;
+use Nelmio\Alice\Throwable\Exception\Generator\Resolver\UnresolvableValueExceptionFactory;
+
+final class FixtureMethodCallReferenceResolver implements ChainableValueResolverInterface, ValueResolverAwareInterface
+{
+    use IsAServiceTrait;
+
+    /**
+     * @var ValueResolverInterface
+     */
+    private $argumentResolver;
+
+    public function __construct(ValueResolverInterface $resolver = null)
+    {
+        $this->argumentResolver = $resolver;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function withValueResolver(ValueResolverInterface $resolver): self
+    {
+        return new self($resolver);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function canResolve(ValueInterface $value): bool
+    {
+        return $value instanceof FixtureMethodCallValue;
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @param FixtureMethodCallValue $value
+     *
+     * @throws NoSuchPropertyException
+     * @throws UnresolvableValueException
+     */
+    public function resolve(
+        ValueInterface $value,
+        FixtureInterface $fixture,
+        ResolvedFixtureSet $fixtureSet,
+        array $scope,
+        GenerationContext $context
+    ): ResolvedValueWithFixtureSet
+    {
+        if (null === $this->argumentResolver) {
+            throw ResolverNotFoundExceptionFactory::createUnexpectedCall(__METHOD__);
+        }
+
+        $functionCall = $value->getFunctionCall();
+        $arguments = $functionCall->getArguments();
+        foreach ($arguments as $index => $argument) {
+            if ($argument instanceof ValueInterface) {
+                $resolvedSet = $this->argumentResolver->resolve($argument, $fixture, $fixtureSet, $scope, $context);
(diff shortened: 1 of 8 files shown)


tests/Generator/Resolver/Value/Chainable/FixtureMethodCallReferenceResolverTest.php
```diff
@@ -0,0 +1,215 @@
...
+        $context->markIsResolvingFixture('foo');
+
+        $valueResolverContext = new GenerationContext();
+        $valueResolverContext->markIsResolvingFixture('foo');
+        $valueResolverContext->markAsNeedsCompleteGeneration();
+
+        $dummyProphecy = $this->prophesize(DummyWithGetter::class);
+        $dummyProphecy->getFoo('resolved_argument')
+            ->shouldBeCalled()
+            ->willReturn('resolved_value');
```
theofidry: the semi-column should be on the new line (yeah I know not everyone likes it, but let's be consistent :P)

tests/Generator/Resolver/Value/Chainable/FixtureMethodCallReferenceResolverTest.php
```diff
@@ -0,0 +1,215 @@
...
+                    'resolved_argument',
+                    $newSet = ResolvedFixtureSetFactory::create(new ParameterBag(['ping' => 'pong']))
+                )
+            )
+        ;
+        $valueResolverProphecy
+            ->resolve($reference, $fixture, $newSet, $scope, $valueResolverContext)
+            ->willReturn(
+                new ResolvedValueWithFixtureSet(
+                    $instance = $dummyProphecy->reveal(),
```
theofidry: I prefer the:

```php
/** @var DummyWithGetter $dummy */
$dummy = $dummyProphecy->reveal();
```

convention as you don't have typehint warnings for the services with static analysis tools

src/Generator/Resolver/Value/Chainable/FixtureMethodCallReferenceResolver.php
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        $context->markAsNeedsCompleteGeneration();
+        $fixtureReferenceResult = $this->argumentResolver->resolve($value->getReference(), $fixture, $fixtureSet, $scope, $context);
+        $context->unmarkAsNeedsCompleteGeneration();
+
+        /** @var ResolvedFixtureSet $fixtureSet */
+        list($instance, $fixtureSet) = [$fixtureReferenceResult->getValue(), $fixtureReferenceResult->getSet()];
+
+        if (! method_exists($instance, $functionCall->getName())) {
```
theofidry: the project convention is to use `false ===` instead

theofidry: I think this check could be avoided or rather done in the catch block below. That way you only check when there's an error hence there is not performance impact for the regular user, and the for the one who did a mistake it stays as user friendly

tests/Generator/Resolver/Value/Chainable/FixtureMethodCallReferenceResolverTest.php
```diff
@@ -0,0 +1,215 @@
...
+        $resolver = new FixtureMethodCallReferenceResolver($valueResolver);
+        $actual = $resolver->resolve($value, $fixture, $set, $scope, $context);
+
+        $this->assertEquals($expected, $actual);
+
+        $valueResolverProphecy->resolve(Argument::cetera())->shouldHaveBeenCalledTimes(2);
+        $dummyProphecy->getFoo(Argument::cetera())->shouldHaveBeenCalledTimes(1);
+    }
+
+    public function testCatchesThrowablesToThrowResolverException()
```
theofidry: `testThrowsAResolverExceptionOnError`? (error to be taken in the english sense, not the `\Error` object)

tests/Generator/Resolver/Value/Chainable/FixtureMethodCallReferenceResolverTest.php
```diff
@@ -0,0 +1,215 @@
...
+            $resolver->resolve($value, new FakeFixture(), $set, [], new GenerationContext());
+
+            $this->fail('Expected exception to be thrown.');
+        } catch (UnresolvableValueException $exception) {
+            $this->assertEquals(
+                'Could not resolve value "mutable->getFoo()".',
+                $exception->getMessage()
+            );
+            $this->assertEquals(0, $exception->getCode());
+            $this->assertEquals($ex, $exception->getPrevious());
```
theofidry: `$this->assertSame()`

tests/Generator/Resolver/Value/Chainable/FixtureMethodCallReferenceResolverTest.php
```diff
@@ -0,0 +1,215 @@
...
+    {
+        try {
+            $value = new FixtureMethodCallValue(
+                $reference = new MutableValue(new DummyWithGetter()),
+                $functionCall = new FunctionCallValue('getFoo')
+            );
+            $set = ResolvedFixtureSetFactory::create(new ParameterBag(['foo' => 'bar']));
+
+            $dummyProphecy = $this->prophesize(DummyWithGetter::class);
+            $dummyProphecy->getFoo()->willThrow($ex = new \Exception('Inner exception'));
```
theofidry: ~~`$ex = new \Exception('Inner exception')`~~ `$error = new \Error()`

tests/Generator/Resolver/Value/Chainable/FixtureMethodCallReferenceResolverTest.php
```diff
@@ -0,0 +1,215 @@
...
+                new GenerationContext()
+            );
+
+            $this->fail('Expected exception to be thrown.');
+        } catch (NoSuchMethodException $exception) {
+            $this->assertEquals(
+                'Could not find the method "getNonExistent" of the object "dummy" (class: Dummy).',
+                $exception->getMessage()
+            );
+            $this->assertEquals(0, $exception->getCode());
```
theofidry: there is not previous in that case? (if so I think a `$this->assertNull($exception->getPrevious))` would be appropriate

tests/Loader/LoaderIntegrationTest.php
```diff
@@ -2575,5 +2575,29 @@ public function provideFixturesToGenerate()
                 ],
             ],
         ];
+
+        yield 'method call reference value' => [
```
theofidry: could you check if there is a test for a method taking a parameter? If not I think it's worth adding as well in that PR

tests/Generator/Resolver/Value/Chainable/FixtureMethodCallReferenceResolverTest.php
```diff
@@ -0,0 +1,220 @@
...
+        try {
+            $value = new FixtureMethodCallValue(
+                $reference = new MutableValue(new DummyWithGetter()),
+                $functionCall = new FunctionCallValue('getFoo')
+            );
+            $set = ResolvedFixtureSetFactory::create(new ParameterBag(['foo' => 'bar']));
+
+            $error = new \Error();
+            $dummyProphecy = $this->prophesize(DummyWithGetter::class);
+            $dummyProphecy->getFoo()->will(function () use ($error) {
```
alekitto: Can't use `willThrow` here as Prophecy `ThrowPromise` cannot be constructed with Errors

tests/Generator/Resolver/Value/Chainable/FixtureMethodCallReferenceResolverTest.php
alekitto: 👍 added

tests/Loader/LoaderIntegrationTest.php
alekitto: Unit test with argument resolution was present.
I've added a test here with a method taking an argument that is a reference to another object.

theofidry:
Thanks you @alekitto

PR #671, opened 2017-02-08
Fix edge case on OptionalValueResolver

### LINKED ISSUE #670: Flaky test for OptionalValue
[This job](https://travis-ci.org/nelmio/alice/jobs/196748892) fails because of `testLoadOptionalValues` assertion failure.

The problem seems located in the `OptionalValueResolver` at line 89:

```
$realValue = (mt_rand(0, 100) <= $quantifier)
    ? $value->getFirstMember()
    : $value->getSecondMember()
```

This will result in the first value to be returned even if the quantifier is 0, in case `mt_rand` generates 0.
I think that the check should be changed in `(mt_rand(0, 99) < $quantifier)`

alekitto:
Fixes #670

### DIFF
--- a/src/Generator/Resolver/Value/Chainable/OptionalValueResolver.php
+++ b/src/Generator/Resolver/Value/Chainable/OptionalValueResolver.php
@@ -86,7 +86,7 @@ final class OptionalValueResolver implements ChainableValueResolverInterface, Va
             }
         }
 
-        $realValue = (mt_rand(0, 100) <= $quantifier)
+        $realValue = (mt_rand(0, 99) < $quantifier)
             ? $value->getFirstMember()
             : $value->getSecondMember()
         ;


alekitto:
Tests are failing on sf3.3-dev because of this: https://github.com/symfony/symfony/commit/bcd897cfd4f92cdc085e2747175f9aff3a76c755 with no stable `psr/container` release available

theofidry:
Would it be possible to add a test for it (the 2 extremums and one fixed non-extrema value)? You could mock `mt_rand`via https://github.com/php-mock/php-mock or even manually cf. http://marcelog.github.io/articles/php_mock_global_functions_for_unit_tests_with_phpunit.html

alekitto:
Was askin' myself how to test it, but didn't know about the `php-mock` library. Thank you for the hint!

theofidry:
Thanks :)

PR #673, opened 2017-02-13
Add covers annotations validator

### LINKED ISSUE #650: Fix wrong covers annotation
Having a wrong `@covers` tags should make the build fail and a check should be done before running the tests

alekitto:
Use [ockcyp/covers-validator](https://github.com/oradwell/covers-validator) to check all the `@covers` annotations before running the test suite and fixed a wrong annotation.

Fixes #650


theofidry:
Nice! Have you tried to see if it makes the tests fail as expected without the fix?

alekitto:
Yes, the build fails at install stage if a covers annotation is invalid.

PR #679, opened 2017-02-23
Increase min version of symfony/property-access

alekitto:
`symfony/property-access` dependency version should be increased to at least 2.3, since `PropertyAccess::createPropertyAccessorBuilder()` (used in NativeLoader) does not exists in 2.2.

Maybe we should add a `--prefer-lowest` build in test matrix?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,6 @@ matrix:
 
 before_install:
   - set -eo pipefail
-  - curl -O -L https://github.com/oradwell/covers-validator/releases/download/v0.5.0/covers-validator.phar
   - echo "memory_limit=-1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
   - phpenv config-rm xdebug.ini || true
   - |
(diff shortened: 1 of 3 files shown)


theofidry:
>Maybe we should add a --prefer-lowest build in test matrix?

Definitely. I should have done that long ago already. Do you mind adding one within this PR? You should have some examples in https://github.com/theofidry/AliceDataFixtures or https://github.com/hautelook/AliceBundle

alekitto:
I think these are the minimum versions to run the test suite correctly:
`fzaninotto/faker`: 1.4 as `numberBetween` does not exists in 1.3 or less
`phpspec/prophecy`: 1.6 to support PHP7
`symfony/console`: 2.7 as it is needed by covers validator command (but this should be handled by `ockcyp/covers-validator` package)

theofidry:
I'm cool with that. Although I would be tempted to say the dependencies of ockcyp/covers-validator should not matter to us and ideally it should be included as a phar but well, let's be pragmatic.

alekitto:
Good point though

.travis.yml
```diff
@@ -28,6 +30,7 @@ matrix:
 
 before_install:
   - set -eo pipefail
+  - curl -O -L https://github.com/oradwell/covers-validator/releases/download/v0.5.0/covers-validator.phar
```
theofidry: I must say I'm not a big fan of that solution for two reasons;

- It requires a download on each build. You could move it in a special directory which is cached to avoid that but then comes the problem of managing that cache: when do you invalidate it i.e. when do you download again that file?
- It is no longer tracked in the project, i.e. a contributor (myself included) no longer knows that this tool is being used before submitting a build.

Both can be acceptable tradeoff to be fair, but I think an easier solution would be to add `oradwell/covers-validator` in a dedicated bamarni namespace (as it is done for symfony and the mutation testing). To be honest I think this is what should have been done in the first place, my bad for not noticing and flagging this in the PR introducing `oradwell/covers-validator`.

theofidry:
Thanks :)

PR #681, opened 2017-03-07
Make NativeLoader non-final

dantleech:
As discussed, this PR makes the NativeLoader non-final in order that it can be extended.


dantleech:
Shall we rename `protected createBuiltIn.*` to `protected create.*` instead?

theofidry:
>Shall we rename protected createBuiltIn.* to protected create.* instead?

Yeah, makes sense. Let's do it in this PR as well

dantleech:
Done - I didn't realise that you could call `$loader->getXxx` externally, so its a "BC" break.

theofidry:
Cheers

PR #683, opened 2017-03-08
add support of embedded couchdb document (couchdb-odm)

zephyr-thomas:
couchdb-odm uses `isEmbeddedDocument` attribute instead of `isEmbeddedClass`.


theofidry:
Nice! However as there is no integration tests for the persistence layer, make sure to test it properly on your end please :)

Also any interest in porting this to Alice 3.x? i.e. adding an adapter in https://github.com/theofidry/AliceDataFixtures.

PR #685, opened 2017-03-14
Nested PHP statement failing test

dantleech:
Discovered this issue when trying to do `<json_encode([])>`, which issues:

```
Nelmio\Alice\Throwable\Exception\FixtureBuilder\ExpressionLanguage\MalformedFunctionException: The value "<(implode(" "" contains an unclosed function.
```

The example in the test is from the documentation: https://github.com/nelmio/alice/blob/master/doc/advanced-guide.md#functions

### DIFF
--- a/tests/Loader/LoaderIntegrationTest.php
+++ b/tests/Loader/LoaderIntegrationTest.php
@@ -1929,6 +1929,24 @@ class LoaderIntegrationTest extends \PHPUnit_Framework_TestCase
             ],
         ];
 
+        yield '[function] call nested PHP native function' => [
+            [
+                \stdClass::class => [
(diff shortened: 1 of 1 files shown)


dantleech:
Note that `foobar: '<strtolower("hello")>'` works, while `foobar: '<json_encode("hello")>'` does not.

theofidry:
>Note that `foobar: '<strtolower("hello")>'` works, while `foobar: '<json_encode("hello")>'` does not.

🤔  Why is that? because of the underscore?

dantleech:
> :thinking: Why is that? because of the underscore?

don't think so - the failing test has no underscores. will try and investigate

PR #705, opened 2017-03-22
Prepare for PHPUnit 6

hboomsma:
After these changes the project is compatible with PHPUnit 6.0.
The symfony/phpunit-bridge however is still not compatible, so
the version requirement in the composer.json is not elevated to
^6.0 yet. As soon as symfony fixes their bridge this can be done.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -10,11 +10,12 @@
   -->
 
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/5.2/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.0/phpunit.xsd"
          backupGlobals="false"
          bootstrap="vendor/autoload.php"
          colors="true"
          enforceTimeLimit="true"
+         beStrictAboutTestsThatDoNotTestAnything="false"
 >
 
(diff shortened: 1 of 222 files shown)


phpunit.xml.dist
```diff
@@ -10,11 +10,12 @@
   -->
 
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/5.2/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.0/phpunit.xsd"
          backupGlobals="false"
          bootstrap="vendor/autoload.php"
          colors="true"
          enforceTimeLimit="true"
+         beStrictAboutTestsThatDoNotTestAnything="false"
```
theofidry: is that change necessary?

hboomsma: Not yet, but in PHPUnit 6.0 you will get a lot of risky tests and exit code 1 when you have no explicit asserts in them. This maintains 5.x behaviour for now.

But since I did not update the constraint yet, it is not strictly necessary.

hboomsma: I can remove it, with no problem, just let me know what is preferable.

theofidry: Ah I see it becomes strict by default in 6.0. Fine then :)

theofidry:
Thanks!

PR #707, opened 2017-03-23
Prepare for PHPUnit 6

hboomsma:
* Made PHPUnit a dev requirement (no dependencies on PHPUnit in the src folder)
* Allow for installing PHPUnit 6
* As Soon as the Symfony PHPunit bridge is released you'll get 6 on PHP 7.0
* Made @covers use FQCN's (as is the case in master)
* Removed old mock framework calls
* Added testcases for the only 2 tests marked as Risky by PHPUnit 6

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,14 +21,14 @@
     "require": {
         "php": "^5.6||^7.0",
         "fzaninotto/faker": "^1.5",
-        "symfony/yaml": "^2.0||^3.0",
-        "phpunit/phpunit": "^5.6"
+        "symfony/yaml": "^2.0||^3.0"
     },
     "require-dev": {
         "doctrine/common": "^2.3",
         "symfony/phpunit-bridge": "^3.0",
         "symfony/property-access": "^2.2||^3.0",
-        "phpspec/prophecy": "^1.5.0"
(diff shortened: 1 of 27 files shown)


theofidry:
>Made PHPUnit a dev requirement (no dependencies on PHPUnit in the src folder)

That was clearly a mistake. As nelmio/alice is usually a dev dependency hopefully this will not have caused much harm though.

Thanks :)

stof:
@theofidry can you make a new 2.3.1 release to avoid forcing installing PHPUnit when using alice ?

theofidry:
I'll do it. If it's not done in the next 2h ping me on Slack

theofidry:
done*

stof:
thanks

stof:
actually, Packagist does not yet know it, while it found 3.0 beta 5.

@Seldaek can you check whether Packagist triggered an error (or just force an update) ?

theofidry:
Damn, it's the second time it's happening. @Seldaek could that be the hooks not working correctly? Otherwise maybe add me on packagist as well so that I can do a force update.

theofidry:
Should be ok now @stof :)

stof:
thanks

PR #708, opened 2017-03-23
Prepare for PHP7.2

hboomsma:
* Run composer with --ignore-platform-reqs for nightly
* Run composer with weak deprecation warnings (PHPUnint still uses each())
* Rewrote usage of each() (which is deprecated in PHP7.2)
* `preg_match` returns `NULL` instead of `''` since PHP7.2

### DIFF
--- a/src/Nelmio/Alice/Fixtures/Fixture.php
+++ b/src/Nelmio/Alice/Fixtures/Fixture.php
@@ -368,19 +368,25 @@ class Fixture
             throw new \UnexpectedValueException("The __construct call in object '{$this}' must be defined as an array of arguments or false to bypass it");
         }
 
-        list($method, $args) = each($constructorValue);
-        if ($method !== 0) {
-            if (!is_callable([$this->class, $method])) {
-                throw new \UnexpectedValueException("Cannot call static method '{$method}' on class '{$this->class}' as a constructor for object '{$this}'");
-            }
-            if (!is_array($args)) {
-                throw new \UnexpectedValueException("The static '{$method}' call in object '{$this}' must be given an array");
+        foreach ($constructorValue as $method => $args) {
+            if ($method !== 0) {
+                if (!is_callable([$this->class, $method])) {
+                    throw new \UnexpectedValueException(
+                        "Cannot call static method '{$method}' on class '{$this->class}' as a constructor for object '{$this}'"
+                    );
+                }
+
+                if (!is_array($args)) {
+                    throw new \UnexpectedValueException(
+                        "The static '{$method}' call in object '{$this}' must be given an array"
+                    );
+                }
+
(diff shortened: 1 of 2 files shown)


src/Nelmio/Alice/Instances/Processor/Processable.php
```diff
@@ -45,6 +45,12 @@ public function getValue()
     public function valueMatches($regex)
     {
         if (preg_match($regex, $this->value, $matches)) {
+            $matches       = array_map(
```
theofidry: if the preg_match change is deprecated I would not change that yet, it wouldn't be the first time such things are reverted and I didn't see much discussion in https://github.com/php/php-src/pull/1303.

src/Nelmio/Alice/Fixtures/Fixture.php
```diff
@@ -368,19 +368,25 @@ protected function getConstructorComponents()
...
 
-        list($method, $args) = each($constructorValue);
-        if ($method !== 0) {
-            if (!is_callable([$this->class, $method])) {
-                throw new \UnexpectedValueException("Cannot call static method '{$method}' on class '{$this->class}' as a constructor for object '{$this}'");
-            }
-            if (!is_array($args)) {
-                throw new \UnexpectedValueException("The static '{$method}' call in object '{$this}' must be given an array");
+        foreach ($constructorValue as $method => $args) {
+
```
theofidry: useless blank line

src/Nelmio/Alice/Fixtures/Fixture.php
```diff
@@ -368,19 +368,25 @@ protected function getConstructorComponents()
...
+            if ($method !== 0) {
+                if (!is_callable([$this->class, $method])) {
+                    throw new \UnexpectedValueException(
+                        "Cannot call static method '{$method}' on class '{$this->class}' as a constructor for object '{$this}'"
+                    );
+                }
+                if (!is_array($args)) {
```
theofidry: while we are at it let's add a blank line here between the two if :)

src/Nelmio/Alice/Fixtures/Fixture.php
hboomsma: :+1:

src/Nelmio/Alice/Fixtures/Fixture.php
hboomsma: :+1:

src/Nelmio/Alice/Instances/Processor/Processable.php
hboomsma: I'll remove this, it will result in a failing nightly branch for now. When PHP decides to be BC again the branch should turn green (on a rebuild)

PR #712, opened 2017-04-05
[Needs help] Replace in-house lexer by HoaCompiler

### LINKED ISSUE #601: Consider a third-party compiler/lexer
As mentioned in #600, the current lexer/parser of the Expression Language is completely custom. While it does the job, I can't say I'm very proud of the implementation and it's far from my field of expertise. Relying on a third-party library for that task would make sense, maybe:

- [HoaCompiler](https://github.com/hoaproject/Compiler)
- [Symfony Expression Language](http://symfony.com/doc/current/components/expression_language.html) (does not looks suitable for this task though, already too high level)

And eventually other (I didn't take the time to properly look into it).

The goal of this component is to be able to transform values as described in #377. Maybe a more detailed example is the actual integration test of the Expression Language parser: [ParserIntegrationTest](https://github.com/nelmio/alice/blob/master/tests/FixtureBuilder/ExpressionLanguage/Parser/ParserIntegrationTest.php)

theofidry:
Note: this message has been edited to avoid the read of the long discussion of here and the original issue (#601).

Alice ships with an Expression Language, which allows to interpret values such as `@user*` or `<current()>`. This is currently done by an in-house Expression Language which uses a Lexer to tokenize the string input and then a Parser which go through those tokens to transform it in an understandable expression. For example:

input: `'@user*'`

tokens returned by the Lexer:
```php
[
   new Token('@user*', new TokenType(TokenType::WILDCARD_REFERENCE_TYPE)),
]
```

value returned by the Parser:
```php
FixtureMatchReferenceValue::createWildcardReference('user')
```

As explained in greater details [here](https://github.com/nelmio/alice/blob/master/CONTRIBUTING.md#architecture), the fixtures are first build into an understandable structure of objects and then they are evaluated to generate the object we wants.

The plan now is to replace the in-house lexer with [HoaCompiler](https://github.com/hoaproject/Compiler). The current one relies on regexes and different pass to try to tokenize the input and then the parser trying to create objects from it. HoaCompiler however works with a better structure: a "grammar" which is a set of rules on how to parse strings is created and return a list of nodes from it. That approach is way more robust and would allow to avoid a lot of edge cases where regexes are just not the tool for the job.

Implementation wise it's not all too complex. Right now most of the Expression Language is tagged as internal so we can afford BC breaks on that part of the library which gives enough freedom to do that work without the need of another major release.

The current PR provides a start of implementation. The major work to do is to write the grammar which @Hywan already started to match our needs. Once done we can then update the Parser to process those `Node` objects instead of the tokens from the in-house lexer (that part should be relatively easy).

Most of the rules that the Expression Language implements are documented [here](https://github.com/nelmio/alice/blob/master/doc/advanced-guide.md#expression-language-dsl) besides the tests.

### DIFF
--- a/Alice.pp
+++ b/Alice.pp
@@ -1,0 +1,62 @@
+//
+// LEXEMES
+//
+%token true true
+%token false false
+%token null null
+%token escape_token \\
+%token string .+
+
+
+//
+// RULES
+//
+value:
+    string()
+
+string:
+    ::escape_token:: <string> | <escape_token> | <string>
+
+//
+//
+//
+//
+//
+//
+//%skip   space          \s
+//// Scalars.
+//%token  true           true
+//%token  false          false
+//%token  null           null
+//// Strings.
+//%token  quote_         <{        -> string
+//%token  string:string  [^"]+
+//%token  string:_quote  }>        -> default
+//// Objects.
+//%token  brace_         {
+//%token _brace          }
+//// Arrays.
+//%token  bracket_       \[
+//%token _bracket        \]
+//// Rest.
+//%token  colon          :
+//%token  comma          ,
+//%token  number         \d+
+//
+//value:
+//    <true> | <false> | <null> | string() | object() | array() | number()
+//
+//string:
+//    ::quote_:: <string> ::_quote::
+//
+//number:
+//    <number>
+//
+//#object:
+//    ::brace_:: pair() ( ::comma:: pair() )* ::_brace::
+//
+//#pair:
+//    string() ::colon:: value()
+//
+//#array:
+//    ::bracket_:: value() ( ::comma:: value() )* ::_bracket::
--- a/Alice.pp
+++ b/Alice.pp
@@ -1,62 +1,163 @@
-//
-// LEXEMES
-//
-%token true true
-%token false false
-%token null null
-%token escape_token \\
-%token string .+
+// All whitespaces matter, except the trailing ones.
+%skip  trailing_whitespaces    \s+$
+
+// An opening chevron must not be escaped by a backslash.
+// Matching an opening chevron changes the namespace from `default` to
+// `parameter`.
+%token opening_chevron    (?<!\\)<    -> parameter
+
+// All whitespaces.
+%skip parameter:whitespaces    \s+
+
+// A closing chevron.
+// Matching an closing chevron changes the namespace from `parameter`
+// to `default`.
+%token parameter:closing_chevron    >    -> __shift__
+
+// A variable opening.
+%token parameter:opening_variable    {    -> variable
+
+// All whitespaces.
+%skip variable:whitespaces    \s+
+
+// An expansion list separator.
+%token variable:comma    ,
+
+// A range separator
+%token variable:range    \.\.
+
+// A range bound.
+%token variable:number    [+-]?[0-9]+
+
+// A variable name can be anything except `}`.
+%token variable:name    [_\w][_\w\d]*
+
+// A variable closing.
+%token variable:closing_variable    }    -> __shift__
+
+// Opening parenthesis.
+%token parameter:opening_parenthesis    \(
+
+// Closing parenthesis.
+%token parameter:closing_parenthesis    \)
+
+// Constant string.
+%token parameter:string    ("|')(.*?)(?<!\\)\1
+
+// A comma used to separate items in a list.
+%token parameter:comma    ,
+
+// A variable or a function name.
+%token parameter:name    [_\w][_\w\d]*
 
 
-//
-// RULES
-//
-value:
-    string()
+// A reference is prefixed by an `@`.
+%token at    @    -> reference
 
-string:
-    ::escape_token:: <string> | <escape_token> | <string>
+// A star is a glob operator.
+%token reference:star    \*    -> __shift__
 
+// A left curly bracket introduces an expansion.
+%token reference:opening_expansion    {    -> expansion
+
+// All whitespaces.
+%skip expansion:whitespaces    \s+
+
+// A number can be signed or not.
+%token expansion:number    [-+]?[0-9]+
+
+// A range is represented by two dots.
+%token expansion:range    \.\.
+
+// A comma is the name separator.
+%token expansion:comma    ,
+
+// A reference expansion name is just like a reference constant name.
+%token expansion:name    [_\w][_\w\d]*
+
+// A right curly bracket closes an expansion.
+%token expansion:closing_expansion    }    -> __shift__ * 2
+
+// A reference name is dynamic if some parts of its name are known at runtime.
+%token reference:dynamic_name    [_\w][_\w\d]*(?=[\*\{])
+
+// A constant reference name is not a dynamic reference name.
+%token reference:constant_name    [_\w][_\w\d]*    ->    __shift__
+
+// Anything is a little bit tricky because it must stop on an
+// unescaped opening chevron. Thus:
+//      .+
+// is wrong because it is greedy. It must be lazy, so:
+//     .+?
 //
+// However, it does not take into account the opening chevron. Thus:
+//     .+?(?=<)
 //
+// This is valid but it does not take into account that the opening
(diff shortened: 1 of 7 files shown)


Hywan:
Hello @theofidry,

Sorry for the late reply. I have a new job, not really free for open source right now. I will try to find time. I am not forgetting you at all :-). Please, feel free to ping me at anytime if I stay silent for too long.

Thanks!

theofidry:
No worries @Hywan, I'm not that free either lately :P

Hywan:
Hello @theofidry :-),

I have a little bit more free time right now. And as promised, I will help on this PR.

So here is a little bit of vocabulary to speak the same language:

* The language we are defining exist in 2 forms: Textual, and in-memory PHP object,
* The goal of the “parser” is to transform the text into in-memory PHP object,
* The textual form is called the _input_,
* There are 2 analysers that form the “parser”:
  1. _Lexical analyser_, also called the _lexer_ that splits a the input into a sequence of _lexemes_ (aka _tokens_),
  2. _Syntactic analyser_, also called the _parser_ that checks if the tokens are correctly ordered by deriving the sequence of tokens based on rules (defined in a grammar),
* If both analysers succeed, then an _Abstract Syntax Tree_ can be produced. Such a tree can be visited easily to check more constraints,
* An AST can be transformed into a _object model_: An enhanced API on top of your language, like the DOM for HTML for instance. The object model is your in-memory PHP objects for instance.

This is a very classical front-end and middle-end compilation workflow. There is other approach of course, but let's stick on this one.

<hr />

Based on that, what do you want?

> So my plan is to just replace the Lexer with HoaCompiler

You want to replace the lexer of Symfony Expression Language by `hoa/compiler`'s. This is possible, but there is a challenge: The token format used by `hoa/compiler` might not be understood by the parser of Symfony Expression Language.

> the Parser which transforms tokens into VOs

What is VO?

<hr />

I have started to implement your DSL grammar. However, before going further, I need clarifications here about your exact needs. If you already have an object model defined, then we can “map”/transform an AST to this object model, and everything else will roll.

theofidry:
Thanks for looking into it @Hywan :)

So basically right now we have a very simple [ParserInterface](https://github.com/nelmio/alice/blob/master/src/FixtureBuilder/ExpressionLanguage/ParserInterface.php):

```php
interface ParserInterface
{
    /**
     * Parses a value, e.g. 'foo' or '$username' to determine if is a regular value (like 'foo') or is a value that
     * must be processed (like '$username'). If the value must be processed, it will be parsed to generate a value (a
     * ValueInterface instance) ready for processing.
     *
     * @param string $value
     *
     * @throws ExpressionLanguageParseThrowable
     *
     * @return ValueInterface|string|array
     */
    public function parse(string $value);
}
```

which is what I called "Expression Language" although it's not exactly one and it's not Symfony one either. _VO_ stands for _Value Object_, which here are `ValueInterface` objects mainly.

So right now that parser is broken down in two parts:

- The [Lexer](https://github.com/nelmio/alice/blob/master/src/FixtureBuilder/ExpressionLanguage/LexerInterface.php) for which the current behaviour is defined [here](https://github.com/nelmio/alice/blob/master/tests/FixtureBuilder/ExpressionLanguage/Lexer/LexerIntegrationTest.php). Basically it's an in-house one that parses the string with regexes and generates a list of tokens from it
- The [Parser](https://github.com/nelmio/alice/blob/master/src/FixtureBuilder/ExpressionLanguage/Parser/SimpleParser.php) itself which decorates the lexer to tokenize the input and then process the tokens to generate the `ValueInterface` objects

The part I'm trying to get rid off by using HoaCompiler is the Lexer to replace this custom regex-based Lexer and tokens. One done the Parser needs to be adapted to consume the generated AST instead of tokens but I expect that to be rather trivial.

The main issue I would say is the grammar itself. Alice DSL is described [here](https://github.com/nelmio/alice/blob/master/doc/advanced-guide.md#expression-language-dsl) but it's something that has been defined over time and grew organically, so I wouldn't be surprised if there's still a few edge cases where things are ambiguous. (But that's also a reason to move to HoaCompiler: it would expose such cases once for all)

Basically what I did in this PR is the starting point: replacing the lexer itself, i.e. starting to define the grammar to transform the input string into an AST to be consumed later.

Let me know if there's one point that is still a bit unclear

Hywan:
OK, let's give it a try :-).

Hywan:
I have few questions so far (I will have more later 😉):

* Can a reference be inside a parameter?
* Can a reference be surrounded by “anything” just like parameters? I mean: Parameters are defined as: `anything()? (parameter() anything()?)*` mostly, like `foo <{bar}> baz`; is it the same for references, like `foo @bar baz`?
* In the given example `@user<numberBetween(1,20)>`, could we have recursion here, something like `@foo<bar(<baz>)>`, I hope no because it's a little bit harder (because of the escaping system and the “anything” in parameters).

theofidry:
>Can a reference be inside a parameter?

From the doc: 

>'<{X}>' where X is evaluated and can be anything.

and I can see a `'<{param<{param2}>}>'` so I would say _yes_. However this looks like a good way to shoot yourself in the foot, so I would say: if it's not too hard, then yes and hope people don't abuse of it. If not, I think restricting it to nested parameters is good enough (and I would prefer that tbh)

> Can a reference be surrounded by “anything” just like parameters? I mean: Parameters are defined as: `anything()? (parameter() anything()?)*` mostly, like `foo <{bar}> baz`; is it the same for references, like `foo @bar baz`?

Yes. Also as you can guess, `foo @bar baz` will fail unless the object referenced by `@bar` is stringable. There is some annoying cases where it's ambiguous and I was thinking of using `{}` for delimiters, e.g. `foo @{bar->foo}baz`, but couldn't find an easy way to implement it with regexes.

> In the given example `@user<numberBetween(1,20)>`, could we have recursion here, something like `@foo<bar(<baz>)>`, I hope no because it's a little bit harder (because of the escaping system and the “anything” in parameters).

That's a risk we have with moving to HoaCompiler I guess as before the expression would have likely to be evaluated wrong and been considered invalid (simply because of the implementation limitations not due to an invalid syntax). If we identify such cases, I think it depends of what we have as there's multiple actions that can be taken: handle it in the grammar for that not happening or ensuring the system bails out in a graceful manner so that the user can easily identify the issue and fix it.

Hywan:
OK so basically, we can have any strings with the Alice language inside, is that correct? So `… alice …` where `…` are anything and `alice` is an Alice construction, like a parameter, a reference etc. Is it correct? And all Alice constructions can be recursive?

theofidry:
Yeah, although I don't really see a recursion case I guess in theory that would be possible.

There's a few exceptions though:

- [Identity](https://github.com/nelmio/alice/blob/master/doc/advanced-guide.md#identity) where the content should remain as a string and is lated evaluated with `eval()`
- [Optional](https://github.com/nelmio/alice/blob/master/doc/advanced-guide.md#optional)

Hywan:
Back from vacations with more free time. Work again on this issue :-).

theofidry:
@Hywan I've updated the PR based on the latest state of the work you pushed of the GitLab repository.

You should see the `LexerIntegrationTest` for which we can now replace the array of tokens of the previous system by the proper expected `TreeNode`. It's not the only thing that needs to be updated but that's the very first step, the next one being updating the Parser accordingly. But at least now anyone who which to purse it just need to run

```
bin/phpunit tests/FixtureBuilder/ExpressionLanguage/Lexer/LexerIntegrationTest.php
```

Without any more faff

Hywan:
Thanks! I appreciate your help :-).

cussack:
@theofidry @Hywan is this still the "hot" PR or was this topic dropped or superseded by something else?

theofidry:
No the topic and PR are still relevant. The branch might be ohutdated but
the rebase should be relatively easy.

Hywan:
I've absent for many months, but now I'm back, and I'm catching up everything. This project is down on my todo list, but the list is smaller every day ;-).

theofidry:
I'll be closing this one since the HoaCompiler has been archived.

PR #714, opened 2017-04-14
Fix error on static factory with associative array

ogizanagi:
I don't know if this is something you want to officially support, but at least it used to work in previous versions ^^

```php
User::class => [
    'user_1' => [
        '__construct' => [
            'create' => [
                'username' => 'john.doe',
                'firstname' => 'John',
                'lastname' => 'Doe',
                // ...
            ],
        ],
    ],
],
```

but since beta4, it'll produce the following error:

> [Error]
  Cannot unpack array with string keys

### DIFF
--- a/src/Bridge/Symfony/Resources/config/fixture_builder/denormalizer/fixture.xml
+++ b/src/Bridge/Symfony/Resources/config/fixture_builder/denormalizer/fixture.xml
@@ -92,6 +92,7 @@
         <service id="nelmio_alice.fixture_builder.denormalizer.fixture.specs.constructor.constructor_with_caller_denormalizer"
                  class="Nelmio\Alice\FixtureBuilder\Denormalizer\Fixture\SpecificationBagDenormalizer\Constructor\ConstructorWithCallerDenormalizer">
             <argument type="service" id="nelmio_alice.fixture_builder.denormalizer.fixture.specs.constructor.simple_constructor_denormalizer" />
+            <argument type="service" id="nelmio_alice.fixture_builder.denormalizer.fixture.specs.arguments.simple_arguments_denormalizer" />
         </service>
 
         <service id="nelmio_alice.fixture_builder.denormalizer.fixture.specs.constructor.simple_constructor_denormalizer"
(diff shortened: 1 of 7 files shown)


theofidry:
Makes sense, I simply overlooked it when I did https://github.com/nelmio/alice/pull/698. Do you mind adding some tests in `LoaderIntegrationTest` as well? You should have some examples in the linked PR.

ogizanagi:
I tried adding one, but it fails with a `Nelmio\Alice\Throwable\Exception\FixtureBuilder\Denormalizer\UnexpectedValueException: Could not denormalize the given constructor` exception :/

I don't really get what's happening here, also considering the fix is working properly on my own project and beta5. 😕

ogizanagi:
I managed in a8278b410cb24b67a142205656df8026d569e4fd to fix the tests by injecting & using the argument denormalizer in `ConstructorWithCallerDenormalizer`, but I don't know these changes are legit.

theofidry:
Hm no that doesn't seem right: alice is broken down in 2 steps: creating the fixture objects and generating the actual objects. The first part is supposed to be idempotent and immutable. They shouldn't be mixed.

I cannot really debug it this we, but I would rather say `ArgumentsDenormalizer` or one of it's caller is likely to be the problem

ogizanagi:
Are you sure those changes aren't right? It only updates the first part about generating the fixture objects and to me does not violates the idempotent and immutable rules.
It looked actually weird to try back using the `SimpleConstructorDenormalizer` in the case of a factory method, just to get the arguments fixture objects and build a `MethodCallWithReference` from it.

To me the `ConstructorWithCallerDenormalizer` should:

1. Try using the decorated implementation (here `SimpleConstructorDenormalizer`). If it succeed, just return, if it fails, continue.
2. Resolve the caller reference (here simply the current classname in the scope) and method.
3. Denormalize the arguments. Previous code was using the `simpleConstructorDenormalizer` for this purpose. But why? We just need to denormalize arguments, we know how to build a `MethodCallInterface` instance from previously collected informations.
4. Build and return the `MethodCallWithReference`.

Maybe I'm missing something 😕

theofidry:
Ah, I actually completely misread your change 😅

theofidry:
Thanks @ogizanagi :)

PR #715, opened 2017-04-14
Fix TypeError on true === needsCompleteGeneration() in ObjectGenerator

DZunke:
There was a TypeError  in the SimpleObjectGenerator. If ```false === $context->needsCompleteGeneration()``` was ```false``` the Objects did not get returned but at the next line it needed the FixtureSet instead of the Objects, so the Line before to directly get the Objects into ```$fixtureSet``` was wrong. 

Had not taken a deeper look but at the moment i created this fix my fixtures got generated. The two Tests that will fail with this pr were failing before. Sorry i had not extended the tests of SimpleObjectGenerator but while fixing a TypeError the existing Test was to crazy for me 😉 

Origin Exception:

```
TypeError: Argument 2 passed to Nelmio\Alice\Generator\ObjectGenerator\SimpleObjectGenerator::completeObject() must be an instance of Nelmio\Alice\Generator\ResolvedFixtureSet, instance of Nelmio\Alice\ObjectBag given, called in /mnt/c/MyProject/vendor/nelmio/alice/src/Generator/ObjectGenerator/SimpleObjectGenerator.php on line 91
```


theofidry:
Hm interesting, the change seems right but I'm surprised this is not covered by the tests. Would it be possible to have a scenario showing that issue? This should be easily doable by using custom entities and a test case in `LoaderIntegrationTest`

DZunke:
Hey @theofidry i had a look and created a test case that triggered the error. Hopefully the test is what you'd looked for 😄

PR #716, opened 2017-04-14
Add a ReflectionPropertyAccessor

ogizanagi:
I use this accessor mainly for hydratation because my domain objects do not have proper setters for every properties, or in order to set predictable values for those computed directly in the constructor.
However, I wonder if it should be wired by default or not.

### DIFF
--- a/fixtures/Entity/DummyWithPrivateProperty.php
+++ b/fixtures/Entity/DummyWithPrivateProperty.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of the Alice package.
+ *
+ * (c) Nelmio <hello@nelm.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Nelmio\Alice\Entity;
+
+class DummyWithPrivateProperty
+{
+    private static $staticVal;
+    private $val;
+
+    public function __construct($val = null)
+    {
+        $this->val = $val;
+    }
+}
--- a/src/PropertyAccess/ReflectionPropertyAccessor.php
+++ b/src/PropertyAccess/ReflectionPropertyAccessor.php
@@ -1,0 +1,117 @@
+<?php
+
+/*
+ * This file is part of the Alice package.
+ *
+ * (c) Nelmio <hello@nelm.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types = 1);
+
+namespace Nelmio\Alice\PropertyAccess;
+
+use Nelmio\Alice\IsAServiceTrait;
+use Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException;
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+
+/**
+ * Decorator that fallbacks to reflection in case a property cannot be reached another way.
+ */
+final class ReflectionPropertyAccessor implements PropertyAccessorInterface
+{
+    use IsAServiceTrait;
+
+    /**
+     * @var PropertyAccessorInterface
+     */
+    private $decoratedPropertyAccessor;
+
+    public function __construct(PropertyAccessorInterface $decoratedPropertyAccessor)
+    {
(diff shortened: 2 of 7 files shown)


fixtures/Entity/DummyWithPrivateProperty.php
```diff
@@ -0,0 +1,24 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Nelmio\Alice\Entity;
+
+class DummyWithPrivateProperty
+{
+    private $val;
```
theofidry: you should add a test with a private static property

src/PropertyAccess/ReflectionPropertyAccessor.php
```diff
@@ -0,0 +1,110 @@
...
+    public function getValue($objectOrArray, $propertyPath)
+    {
+        try {
+            return $this->decoratedPropertyAccessor->getValue($objectOrArray, $propertyPath);
+        } catch (NoSuchPropertyException $e) {
+            if (!$this->propertyExists($objectOrArray, $propertyPath)) {
+                throw $e;
+            }
+
+            $getPropertyClosure = \Closure::bind(function ($object) use ($propertyPath) {
```
theofidry: (CS)

```php
$getPropertyClosure = \Closure::bind(
    function ($object) use ($propertyPath) {
        return $object->{$propertyPath};
    },
    $objectOrArray,
    $objectOrArray
);
```

src/PropertyAccess/ReflectionPropertyAccessor.php
```diff
@@ -0,0 +1,110 @@
...
+    private function propertyExists($objectOrArray, $propertyPath)
+    {
+        if (!is_object($objectOrArray)) {
+            return false;
+        }
+
+        $class = get_class($objectOrArray);
+        $reflClass = new \ReflectionClass($class);
+
+        return $reflClass->hasProperty($propertyPath);
```
theofidry:
```php
return (new \ReflectionClass($class))->hasProperty($propertyPath);
```

(I would keep the `$class` though)

tests/Loader/LoaderIntegrationTest.php
```diff
@@ -13,6 +13,7 @@
 
 namespace Nelmio\Alice\Loader;
 
+use Nelmio\Alice\Entity\DummyWithPrivateProperty;
```
theofidry: hm, the order doesn't seem right, I really need to add a StyleCI...

fixtures/Entity/DummyWithPrivateProperty.php
ogizanagi: You mean to ensure the accessor never consider static properties? Done (and accessor updated).

ogizanagi:
Changes made. :)

Where would the documentation fit the best in the current documentation? (mention the decoration of the loader property accessor + symfony configuration sample)

theofidry:
I would say [here](https://github.com/nelmio/alice/blob/master/doc/advanced-guide.md#custom-accessor) and maybe mentioning it [there](https://github.com/nelmio/alice/blob/master/doc/relations-handling.md#references) as well:

>You can reference properties reachable through a getter (i.e : @name->property will call $name->getProperty() if property is not public)

tests/PropertyAccess/ReflectionPropertyAccessorTest.php
```diff
@@ -0,0 +1,260 @@
...
+        $accessor = new ReflectionPropertyAccessor($decoratedAccessor);
+
+        $accessor->getValue($object, $property);
+    }
+
+    /**
+     * @expectedException \Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException
+     * @expectedExceptionMessage Cannot read property "staticVal".
+     */
+    public function testThrowsAnOriginalExceptionIfPropertyIsStatic()
```
theofidry: Shouldn't it be able to read it?

ogizanagi: I don't think we should support it (use cases look a bit edgy to me), but you set the rules. 😄 

Actually, I'm not even sure the property accessor component is meant to be used to access any static property.

ogizanagi:
Documentation added.

PR #717, opened 2017-04-15
Fix decorated service name in arguments

ogizanagi:
Oops. :)


theofidry:
Ha, indeed missed that as well! thanks :)

Issue #724, opened 2017-04-28
The ReferenceEscaperLexer will not escape when there is a number before `@`.


alekslauda:
Using emails like `email1@example.com` are not escaped by the ReferenceEscaperLexer.

https://github.com/nelmio/alice/blob/master/src/FixtureBuilder/ExpressionLanguage/Lexer/ReferenceEscaperLexer.php#L47

theofidry:
@alekslauda would you mind opening a PR to fix it? I don't think this should be too hard

theofidry:
Note that it's a known issue and can be fixed by escaping it manually: `email1\@example.com`

theofidry:
I've just tested it on master and seems to work. Not sure if your case is more complex or if it has been fixed in the meantime. If you still encounter the problem please open a new issue with the failing email :)

PR #733, opened 2017-05-10
Fix ReferenceEscaperLexer.

ndyakov:
Escape emails ending with a digit. 
Related to #724.

Also @theofidry Keep in mind that the old Escaper would have escaped nested parameters such as `@user@user1`.

### DIFF
--- a/src/FixtureBuilder/ExpressionLanguage/Lexer/ReferenceEscaperLexer.php
+++ b/src/FixtureBuilder/ExpressionLanguage/Lexer/ReferenceEscaperLexer.php
@@ -44,7 +44,7 @@ final class ReferenceEscaperLexer implements LexerInterface
      */
     public function lex(string $value): array
     {
-        $escapedValue = preg_replace('/((\s|^)[a-zA-Z0-9_.+-]+?[A-Za-z0-9])@/', '$1\\@', $value);
+        $escapedValue = preg_replace('/((\s|^)[^@]?(\p{L}|\p{N}|[+-_.])+?(\p{L}|\p{N}))@/', '$1\\@', $value);
         return $this->lexer->lex($escapedValue);
     }
 }
--- a/src/FixtureBuilder/ExpressionLanguage/Lexer/ReferenceEscaperLexer.php
+++ b/src/FixtureBuilder/ExpressionLanguage/Lexer/ReferenceEscaperLexer.php
@@ -44,8 +44,7 @@ final class ReferenceEscaperLexer implements LexerInterface
      */
     public function lex(string $value): array
     {
-        $escapedValue = preg_replace('/(\\p{L})@/', '$1\\@', $value);
-
+        $escapedValue = preg_replace('/((\s|^)[a-zA-Z0-9_.+-]+?[A-Za-z0-9])@/', '$1\\@', $value);
         return $this->lexer->lex($escapedValue);
     }
 }
--- a/tests/FixtureBuilder/ExpressionLanguage/Lexer/ReferenceEscaperLexerTest.php
+++ b/tests/FixtureBuilder/ExpressionLanguage/Lexer/ReferenceEscaperLexerTest.php
@@ -66,5 +66,7 @@ class ReferenceEscaperLexerTest extends TestCase
         yield 'string with a reference with members' => ['bar @foo baz'];
 
         yield 'reference in a middle of a word' => ['email@example', 'email\\@example'];
+
+        yield 'email ending with a digit' => ['email1@example', 'email1\\@example'];
     }
 }


src/FixtureBuilder/ExpressionLanguage/Lexer/ReferenceEscaperLexer.php
```diff
@@ -44,8 +44,7 @@ public function __construct(LexerInterface $decoratedLexer)
      */
     public function lex(string $value): array
     {
-        $escapedValue = preg_replace('/(\\p{L})@/', '$1\\@', $value);
-
+        $escapedValue = preg_replace('/((\s|^)[a-zA-Z0-9_.+-]+?[A-Za-z0-9])@/', '$1\\@', $value);
```
theofidry: an email could have a non ASCII character, so `a-zA-Z` is wrong as `à` for example won't work, hence `p{L}`

tests/FixtureBuilder/ExpressionLanguage/Lexer/ReferenceEscaperLexerTest.php
```diff
@@ -66,5 +66,7 @@ public function provideValues()
         yield 'string with a reference with members' => ['bar @foo baz'];
 
         yield 'reference in a middle of a word' => ['email@example', 'email\\@example'];
+
+        yield 'email ending with a digit' => ['email1@example', 'email1\\@example'];
```
theofidry: given the issues we had, feels like we should add a few more tests with exotic emails (with non ASCII characters for example)

theofidry:
I think that's fine, it looks like more of and edge case that can very well wait #712

src/FixtureBuilder/ExpressionLanguage/Lexer/ReferenceEscaperLexer.php
ndyakov: @theofidry something like 
`((\s|^)[^@]?(\p{L}|\p{N}|[+-_.])+?(\p{L}|\p{N}))@` should work just fine.
I will test this later.

theofidry:
No worries, but it looks like this causes a slight parsing issue

src/FixtureBuilder/ExpressionLanguage/Lexer/ReferenceEscaperLexer.php
```diff
@@ -44,7 +44,7 @@ public function __construct(LexerInterface $decoratedLexer)
      */
     public function lex(string $value): array
     {
-        $escapedValue = preg_replace('/((\s|^)[a-zA-Z0-9_.+-]+?[A-Za-z0-9])@/', '$1\\@', $value);
+        $escapedValue = preg_replace('/((\s|^)[^@]?(\p{L}|\p{N}|[+-_.])+?(\p{L}|\p{N}))@/', '$1\\@', $value);
```
ndyakov: If the regex was written without the optional `[^@]` it would work fine but only emails with at least 2 characters in the name would be matched.

theofidry: I could be ok with that, but I think we need way more tests regarding emails to be able to tell if it's alright or not. Also correct me if I'm wrong but it's about auto-escaping emails, so for tricky edge cases one could always fallback to manual escaping e.g. `a\@b.c`

ndyakov: @theofidry Agreed as far as writing more tests goes. I would prefer no auto-escaping than "sometimes works" auto-escaping. You have the final say, this looks like it could create problems with time, so maybe remove auto-escaping with the next major release ?

theofidry: I think auto-escaping is fine. Most people use regular emails or faker generated emails, it's ok for the 1-2 people using complete alien emails to not benefit from it

theofidry:
@ndyakov any updates? No pressure if you don't have time, it's just to know if I can count on it to be included in the next RC release or if this will wait later

ndyakov:
Nothing at the moment, sorry.

theofidry:
That's ok :)

theofidry:
Is this still relevant? As mentioned in #724 this now seems to be working fine

ndyakov:
@theofidry since then we have moved away from emails ending with numbers for our fixtures, but I can test this if it is working tomorrow. How do you feel about closing this one and probably open a new one with just few more tests regarding the numbers in emails?

theofidry:
Fair enough :)

PR #734, opened 2017-05-11
Rework callers

### LINKED ISSUE #732: Add support for various type of method calls
The current implementation of `CallerInterface`, `SimpleCaller`, is too simple and is missing a few features it should support:

- Optional calls (need to check if they are well respected for the constructor/factories as well)
- Have an extension point to handle calls differently depending of the flags they have (use case: #731)

theofidry:
The last big piece (#731) should be very easy after that one for the 3.0.0.

- Add support for flags in methods, e.g. optional calls
- Add an extension point for flag handling
- Add support for static or service reference calls

Closes #732

LoaderIntegrationTest is passing but the rest needs more polishing:

- [x] Finish the code (there's some missing bits)
- [x] Add missing tests
- [x] Add doc
- [x] Cleanup


theofidry:
@ogizanagi adding support for immutable functions should be trivial now.

PR #735, opened 2017-05-14
Fix optional callables

DZunke:
The ```SimpleCaller``` had no implementation to handle ```OptionalMethodCall```. With this it is again possible to set a percentage on call configuration.

```php
'foo' => [
    '__calls' => [
        ['setBar (50%?)' => []],
        ['setBaz (50%?)' => []],
   ],
 ]
```


theofidry:
I'm covering that among other things in #734 :) Thanks for the PR though, I doubled check that I didn't miss any tests.

PR #774, opened 2017-07-26
[DX] Properly handle missing specs

curry684:
Coalesce missing specs to an empty array to improve error message.

I've been bitten a few times by typing this:
```
My\Fine\Entity:
  base_entity (template):
    username: '<username()>

  entity{1..10} (extends base_entity): ~
```
Which gives a wonderfully unhelpful error as it should be `[]` instead right now.

> PHP Fatal error:  Uncaught TypeError: Argument 4 passed to Nelmio\Alice\FixtureBuilder\Denormalizer\Fixture\FixtureDenormalizerRegistry::denormalize() must be of the type array, null given, called in /vendor/nelmio/alice/src/FixtureBuilder/Denormalizer/Fixture/SimpleFixtureBagDenormalizer.php on line 70 and defined in /vendor/nelmio/alice/src/FixtureBuilder/Denormalizer/Fixture/FixtureDenormalizerRegistry.php:59

This patch coalesces the `null` into an empty array instead to give the expected behavior.

### DIFF
--- a/src/FixtureBuilder/Denormalizer/Fixture/SimpleFixtureBagDenormalizer.php
+++ b/src/FixtureBuilder/Denormalizer/Fixture/SimpleFixtureBagDenormalizer.php
@@ -70,7 +70,7 @@ final class SimpleFixtureBagDenormalizer implements FixtureBagDenormalizerInterf
                     $fixtures,
                     $fqcn,
                     $reference,
-                    $specs,
+                    $specs ?? [],
                     $flags
                 );
             }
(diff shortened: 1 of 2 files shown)


theofidry:
Do you mind adding a small test for it? I believe a test in LoaderIntegrationTest should be enough

curry684:
Yeah I was too lazy for a full fork and multifile edits are bad in the web UI   😉

theofidry:
Many thanks :)

Issue #645, opened 2016-11-23
randomElement


nlemoine:
Hi,

I'm using the latest version of Alice (master branch) and have an issue with the `randomElement` method. 

```yaml
Test\Entity:
  entity{1..20}:
    some_field: <randomElement(['some value', 'some other value'])>
```

I get this error:
```
Uncaught Nelmio\Alice\Exception\FixtureBuilder\ExpressionLanguage\LexException: Could not lex the value "['some value'". in /path/to/vendor/nelmio/alice/src/Exception/FixtureBuilder/ExpressionLanguage/LexException.php:25
``` 

Am I missing something in array syntax or is it a confirmed bug?

theofidry:
No this should work, I'll look into it

zajca:
Hi, 
I can confirm same problem with static methods returning array.

```
domainType: <randomElement(\MyBundle\Entity\MyEntity::getAvailableDomainTypes())>
```

I take a look to my older projects and with 2.2.2 this was working. Now with v3.0.0-beta.1 it's not working.

theofidry:
@zajca the correct syntax for yours would be `<randomElement(<(\MyBundle\Entity\MyEntity::getAvailableDomainTypes())>)>`, unlike in 2.x, there is a few things that changes in 3.x due to introducing a proper lexer. I wish I could keep the BC at this level, but there is far too much edge cases :/

theofidry:
Duplicate of #663.

rouflak:
@theofidry I'm getting the same issue and could not find any solution in the issue mentioned above. 

Could you please write a brief conclusion to let us know how to deal with arrays and the `randomElement` fixture?

Thank you very much

theofidry:
@rouflak did you retry with the latest version? I believed I fixed that one already.

dominics:
@theofidry confirmed the issue still exists in both beta4, and in 4d47f90371 (latest master at this point). (My test case is like the original issue reporter's, just an array literal, not a static call)

Seems like this isn't a duplicate of #663 as well; that provider was being passed a string, not an array.

dominics:
One workaround is to use an array parameter, which doesn't make the lexer spew:

```
parameters:
  planets:
    - mercury
    - venus
    - earth
    
Type:
  name:
    planet: "<randomElement(<{planets}>)>"
```

I also note there's no test case for a literal array (`"<randomElement(['foo', 'bar'])>"`) in the LexerIntegrationTest.

[...]

PR #775, opened 2017-07-31
#645 fixed array function arguments

marinoska:
fixes parsing arrays as a function argument,
example:
the 
 <randomElement(['facebook','instagram','vkontakte'], 'test', ['lalala'])>
should be processed into three tokens/arguments:
['facebook','instagram','vkontakte']
'test'
['lalala']


theofidry:
Thanks very much! Do you mind adding a test to `LexerIntegrationTest` and `ParserIntegrationTest` as well?

marinoska:
ah... In fact I missed it,  I'll add them later then :)

theofidry:
👌 🎉

PR #784, opened 2017-09-11
Add alternative name for static factory methods

greg0ire:
I failed to find what I was looking for (named constructor), although it
seems to be used in the code. This should help finding the docs.


theofidry:
Cheers :)

PR #789, opened 2017-09-14
Add note about installing vendor binaries

simshaun:
It took me a short bit to figure out why `make test` wasn't working.

I thought the composer-bin-plugin would "just work", but I ended up needing to run `composer bin all update`.


theofidry:
Actually I think the Makefile should be update accordingly rather

PR #793, opened 2017-09-14
Throw exception on invalid property

### LINKED ISSUE #792: TypeError: Argument 3 passed to FixtureDenormalizerRegistry::denormalize()
Hi,

I'm upgrading to v3 and I get a lot of these, not sure it is my fault or a bug:

```
TypeError: Argument 3 passed to Nelmio\Alice\FixtureBuilder\Denormalizer\Fixture\FixtureDenormalizerRegistry::denormalize() must be of the type string, integer given, called in /server/http/vendor/nelmio/alice/src/FixtureBuilder/Denormalizer/Fixture/SimpleFixtureBagDenormalizer.php on line 75
```

`Nelmio\Alice\FixtureBuilder\Denormalizer\Fixture\SimpleFixtureBagDenormalizer::denormalize` receives the following array (excerpt):

```
array(2) {
  ["HelloFresh\RecipeService\Domain\Tag\Tag"]=>
  array(3) {
    [0]=>
    array(2) {
      ["localeData"]=>
      array(2) {
        ["en-US"]=>
        string(15) "@tag1LocaleData"
        ["de-DE"]=>
        string(15) "@tag1LocaleData"
      }
      ["numberOfRecipesByCountry"]=>
      array(1) {
        ["US"]=>
        int(10)
      }
    }
```

Indeed, `$reference` is an integer since the array has no associated keys.

Here follows the fixture definition (excerpt):

```yaml
HelloFresh\RecipeService\Domain\Tag\TagLocaleData:
  tag1LocaleData:
    name: tag1

HelloFresh\RecipeService\Domain\Tag\Tag:
  - localeData:
      en-US: "@tag1LocaleData"
      de-DE: "@tag1LocaleData"
    numberOfRecipesByCountry:
      US: 10
```

Is it mandatory to name fixtures since v3, should `$reference` be casted, or was I totally wrong to do that in the first place?

```php
                $fixtures = $this->fixtureDenormalizer->denormalize(
                    $fixtures,
                    $fqcn,
                    (string) $reference,
                    $specs ?? [],
                    $flags
                );
```

olvlvl:
Hi,

Consider the following bad definition (excerpt):

```yaml
Step:
  st1:
    ingredients:
    - "@i1"
  st2:
    - "@i2"
```

Because I forgot the `ingredients` key on `st2` definition I'm presented with the following error:

```
TypeError: Argument 3 passed to Nelmio\Alice\FixtureBuilder\Denormalizer\Fixture\SpecificationBagDenormalizer\SimpleSpecificationsDenormalizer::denormalizeProperty() must be of the type string, integer given, called in /server/http/vendor/nelmio/alice/src/FixtureBuilder/Denormalizer/Fixture/SpecificationBagDenormalizer/SimpleSpecificationsDenormalizer.php on line 99
```

Which appears like a bug in the software more than anything. I mean the application shouldn't crash on a type error.

With this PR, the following exception would be thrown:

```
InvalidArgumentException: Invalid property name: 0.
```

Granted it doesn't give much context about the origin of the error, but it could be improved.

What do you think?

<hr>

Maintainer edit: Closes #792

### DIFF
--- a/src/FixtureBuilder/Denormalizer/Fixture/SpecificationBagDenormalizer/SimpleSpecificationsDenormalizer.php
+++ b/src/FixtureBuilder/Denormalizer/Fixture/SpecificationBagDenormalizer/SimpleSpecificationsDenormalizer.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Nelmio\Alice\FixtureBuilder\Denormalizer\Fixture\SpecificationBagDenormalizer;
 
+use InvalidArgumentException;
 use Nelmio\Alice\Definition\MethodCall\NoMethodCall;
 use Nelmio\Alice\Definition\MethodCallBag;
 use Nelmio\Alice\Definition\MethodCallInterface;
@@ -62,8 +63,13 @@ final class SimpleSpecificationsDenormalizer implements SpecificationsDenormaliz
         $calls = new MethodCallBag();
 
         foreach ($unparsedSpecs as $unparsedPropertyName => $value) {
-            if (!is_string($unparsedPropertyName)) {
-                throw new \InvalidArgumentException("Invalid property name: $unparsedPropertyName.");
+            if (false === is_string($unparsedPropertyName)) {
+                throw new InvalidArgumentException(
+                    sprintf(
+                        'Invalid property name: %s.',
+                        $unparsedPropertyName
+                    )
+                );
             }
 
             if ('__construct' === $unparsedPropertyName) {
--- a/src/FixtureBuilder/Denormalizer/Fixture/SpecificationBagDenormalizer/SimpleSpecificationsDenormalizer.php
+++ b/src/FixtureBuilder/Denormalizer/Fixture/SpecificationBagDenormalizer/SimpleSpecificationsDenormalizer.php
@@ -62,6 +62,10 @@ final class SimpleSpecificationsDenormalizer implements SpecificationsDenormaliz
         $calls = new MethodCallBag();
 
         foreach ($unparsedSpecs as $unparsedPropertyName => $value) {
+            if (!is_string($unparsedPropertyName)) {
+                throw new \InvalidArgumentException("Invalid property name: $unparsedPropertyName.");
+            }
+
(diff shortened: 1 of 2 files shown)


src/FixtureBuilder/Denormalizer/Fixture/SpecificationBagDenormalizer/SimpleSpecificationsDenormalizer.php
```diff
@@ -62,6 +62,10 @@ public function denormalize(FixtureInterface $scope, FlagParserInterface $parser
         $calls = new MethodCallBag();
 
         foreach ($unparsedSpecs as $unparsedPropertyName => $value) {
+            if (!is_string($unparsedPropertyName)) {
```
theofidry: `false ===`, I know it doesn't matter much but I prefer to keep the codebase somewhat consistent :)

src/FixtureBuilder/Denormalizer/Fixture/SpecificationBagDenormalizer/SimpleSpecificationsDenormalizer.php
```diff
@@ -62,6 +62,10 @@ public function denormalize(FixtureInterface $scope, FlagParserInterface $parser
...
+                throw new \InvalidArgumentException("Invalid property name: $unparsedPropertyName.");
```
theofidry: - import the exception
- use sprintf rather

olvlvl:
> You could however catch that exception in the file loader to enrich the exception to specify which file is buggy.

I might need to create a new exception class for that since `InvalidArgumentException` is too generic to be relied on. What do you think?

theofidry:
Yeah it's probably better, just a simple one extending `\InvalidArgumentException`. The design of exceptions for this library are:

- minimal exceptions
- locked exceptions factories: locked as private, for our use only

So that should give you enough freedom to not worry too much about BC

theofidry:
Coming back on that one, I think it's fine with a `InvalidArgumentException`: it's not an error meant to be catched or anything.

theofidry:
Thanks very much @olvlvl :)

PR #807, opened 2017-10-17
Add project name in trigger error deprecation notices

tifabien:
When we execute a PHPUnit test suite with multiple dependencies we need to dig a bit to understand the origin of the deprecation notice message.

In the same way as symfony, I suggest to add in deprecation notices messages the 'Alice' project name


theofidry:
Good call 👍 . Do you mind checking 3.x for that as well? Although much less, there is one or two deprecation message IIRC.

tifabien:
Of course! I'll have a look tomorrow and create a pr if needed.

PR #810, opened 2017-10-19
Add Project name in trigger error deprecation notices

tifabien:
@theofidry here it is for the master branch :)


theofidry:
Thanks :)

PR #812, opened 2017-10-20
Legacy Testing

benji07:
When i tried to run the test in local i got 3 warning like this one

```
Nelmio\Alice\FixtureBuilder\Denormalizer\Fixture\SpecificationBagDenormalizer\Constructor\LegacyConstructorDenormalizerTest::testIsNotClonable
Using the "Legacy" prefix to mark all tests of a class as legacy is deprecated since version 3.3 and will be removed in 4.0. Use the "@group legacy" notation instead to add the test to the legacy group.
```

So i add the `@group legacy` annotation, i don't know if this is the correct thing to do


theofidry:
I think instead we could simply break the naming convention for the test class, so rename it instead. It has `Legacy` is the name because the tested class has it, not to leverage the PHPUnitBridge...

theofidry:
Closed via #818. Thanks still :)

PR #838, opened 2017-11-19
Coding standard fixes

TomasVotruba:
- PSR2
- unused doc blocks
- sort namespace imports

### DIFF
--- a/src/Bridge/Symfony/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/DependencyInjection/Configuration.php
@@ -93,8 +93,7 @@ final class Configuration implements ConfigurationInterface
                         ->always($this->createPositiveIntegerValidatorClosure())
                     ->end()
                 ->end()
-            ->end()
-        ;
+            ->end();
 
         return $treeBuilder;
     }
--- a/src/Bridge/Symfony/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/DependencyInjection/Configuration.php
@@ -93,7 +93,8 @@ final class Configuration implements ConfigurationInterface
                         ->always($this->createPositiveIntegerValidatorClosure())
                     ->end()
                 ->end()
-            ->end();
+            ->end()
+        ;
 
         return $treeBuilder;
     }
--- a/src/FileLocator/DefaultFileLocator.php
+++ b/src/FileLocator/DefaultFileLocator.php
@@ -56,6 +56,7 @@ final class DefaultFileLocator implements FileLocatorInterface
                 && $file[1] === ':'
                 && ($file[2] === '\\' || $file[2] === '/')
             )
-            || null !== parse_url($file, PHP_URL_SCHEME);
+            || null !== parse_url($file, PHP_URL_SCHEME)
+        ;
     }
 }
--- a/src/FileLocator/DefaultFileLocator.php
+++ b/src/FileLocator/DefaultFileLocator.php
@@ -48,7 +48,7 @@ final class DefaultFileLocator implements FileLocatorInterface
 
     private function isAbsolutePath(string $file): bool
     {
-        return ($file[0] === '/'
+        return $file[0] === '/'
             || $file[0] === '\\'
             || (
                 strlen($file) > 3
@@ -57,6 +57,6 @@ final class DefaultFileLocator implements FileLocatorInterface
                 && ($file[2] === '\\' || $file[2] === '/')
             )
             || null !== parse_url($file, PHP_URL_SCHEME)
-        );
+        ;
     }
 }
--- a/src/FileLocator/DefaultFileLocator.php
+++ b/src/FileLocator/DefaultFileLocator.php
@@ -30,7 +30,7 @@ final class DefaultFileLocator implements FileLocatorInterface
      */
     public function locate(string $name, string $currentPath = null): string
     {
-        if ('' == $name) {
+        if ('' === $name) {
             throw FileNotFoundException::createForEmptyFile();
         }
 
--- a/src/FileLocator/DefaultFileLocator.php
+++ b/src/FileLocator/DefaultFileLocator.php
@@ -56,7 +56,6 @@ final class DefaultFileLocator implements FileLocatorInterface
                 && $file[1] === ':'
                 && ($file[2] === '\\' || $file[2] === '/')
             )
-            || null !== parse_url($file, PHP_URL_SCHEME)
-        ;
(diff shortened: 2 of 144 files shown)


src/Bridge/Symfony/DependencyInjection/Configuration.php
```diff
@@ -93,8 +93,7 @@ function (array $value) {
                         ->always($this->createPositiveIntegerValidatorClosure())
                     ->end()
                 ->end()
-            ->end()
-        ;
+            ->end();
```
theofidry: I'm really not a big fan of that change, I find it much less readable and makes less sense

src/FileLocator/DefaultFileLocator.php
```diff
@@ -48,15 +48,14 @@ public function locate(string $name, string $currentPath = null): string
 
     private function isAbsolutePath(string $file): bool
     {
-        return ($file[0] === '/'
+        return $file[0] === '/'
```
theofidry: I would keep those parenthesis, they are here for readability

src/Generator/ObjectGenerator/CompleteObjectGenerator.php
```diff
@@ -62,7 +62,7 @@ public function generate(
 
     private function isObjectComplete(FixtureInterface $fixture, ObjectInterface $object, GenerationContext $context): bool
     {
-        return (
+        return
```
theofidry: same here: I find it easier to read with the parenthesis

tests/Loader/LoaderIntegrationTest.php
```diff
@@ -591,7 +594,7 @@ public function testTemplatesAreBuildBeforeUsage()
         );
 
         $actual = $this->loader->loadData([
-            \Nelmio\Alice\Entity\DummyWithConstructorAndCallable::class => [
```
theofidry: keeping a FQCN is better here as it is way more easier to follow

src/Bridge/Symfony/DependencyInjection/Configuration.php
TomasVotruba: I just run PSR2 as considered to be standard. 
I think it's better to have consistend code than special rule for better readability and writeability.

src/FileLocator/DefaultFileLocator.php
TomasVotruba: When I read some code in (...) I suppose it's required and has special meaning (otherwise every code or condition would have it). This is not the case.

tests/Loader/LoaderIntegrationTest.php
TomasVotruba: Just this case or in general? I think it's better to use expressive class names, rather then depend on namespace, like `UserRepository` instead of `User` (entity? repository? controller?) and `DummyEntityWithConstructorAndCallable` instead of `DummyWithConstructorAndCallable` (dummy something?).

But I can remove this file from change.

TomasVotruba:
> I'm fine with most of the changes but there is a few rules I'm not fond of.

Awesome, I've replied at those with my reasoning. But all can be removed of course. It's your call.

> Also there is a make cs command using PHP-CS-Fixer, would it be possible to add the missing rules to it?

I'd love to, but it's not possible since it's based on [EasyCodingStandard](https://github.com/Symplify/EasyCodingStandard).

tests/Loader/LoaderIntegrationTest.php
theofidry: I agree with the practice in general, but for this case I find it more harmful than anything. It is easier to follow the FQCN (as you don't have to look up for the use statement) and a lot of classes have similar names and using use statements may require aliases. As it's only for those tests, I don't really want to bother with the names.

TomasVotruba: Allright then.

src/FileLocator/DefaultFileLocator.php
theofidry: I genuinely find it easier to read something like

```php
return (
    'foo' === $foo
    && 'bar === $bar
);
```

than:

```php
return 'foo' === $foo
    && 'bar === $bar;
```

(not the best example as in that case it could be easily inlined though).

That said we might just agree to disagree, I think it's very subjective for that one

src/Bridge/Symfony/DependencyInjection/Configuration.php
theofidry: I may be mistaken but I think that's a PHP-CS-Fixer thing not PSR-2, or then the PSR-2 doc needs to be updated as a few samples are using that CS as well

tests/Loader/LoaderIntegrationTest.php
theofidry: that's specific to that file only for those fixtures though :)

src/Bridge/Symfony/DependencyInjection/Configuration.php
TomasVotruba: I didn't find it in examples (`;` is always preceeded by some char), but you are right, it's in `no_empty_statement [@Symfony]`, my bad.

src/FileLocator/DefaultFileLocator.php
TomasVotruba: As you wish :bowing_man:

src/Bridge/Symfony/DependencyInjection/Configuration.php
TomasVotruba: Well, it actually from PHP_CodeSniffer: `PHP_CodeSniffer\Standards\PSR2\Sniffs\Methods\FunctionCallSignatureSniff.SpaceAfterCloseBracket`


image

tests/Loader/LoaderIntegrationTest.php
TomasVotruba: So this whole file should be reverted or only this single case?

tests/Bridge/Symfony/FixtureBuilder/ExpressionLanguage/Parser/ParserIntegrationTest.php
```diff
@@ -13,6 +13,10 @@
 
 namespace Nelmio\Alice\Bridge\Symfony\FixtureBuilder\ExpressionLanguage\Parser;
 
+<<<<<<< HEAD
```
theofidry: I think there's a little error there :)

theofidry:
Many thanks @TomasVotruba :)

TomasVotruba:
Glad to help :+1:

PR #847, opened 2017-12-18
Autoconfigure custom faker providers

theofidry:
- Update Travis configuration
- Bump some dependencies
- Add a conflict rule to prevent installation on incompatible versions of Symfony
- Faker providers are automatically registered when they extend the Faker base class
- Make the `Faker\Generator` service public to allow automatic service binding

### DIFF
--- a/src/Bridge/Symfony/DependencyInjection/NelmioAliceExtension.php
+++ b/src/Bridge/Symfony/DependencyInjection/NelmioAliceExtension.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Nelmio\Alice\Bridge\Symfony\DependencyInjection;
 
+use Faker\Provider\Base as BaseFakerProvider;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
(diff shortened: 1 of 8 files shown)


Soullivaneuh:
I think you should also add a reference for `Faker\Generator` service for auto-binding.

Class guessing does not work anymore on Symfony 4.x, so I have to do this:

```yaml
services:
    _defaults:
        autowire: true
        autoconfigure: true
        public: false

    Faker\Generator: '@nelmio_alice.faker.generator'
```

theofidry:
Makes sense

src/Bridge/Symfony/DependencyInjection/NelmioAliceExtension.php
```diff
@@ -60,11 +61,15 @@ private function loadServices(ContainerBuilder $container)
...
+
         $finder->files()->in(self::SERVICES_DIR);
+
         foreach ($finder as $file) {
             $loader->load(
                 $file->getRelativePathname()
             );
         }
+
+        $container->registerForAutoconfiguration(BaseFakerProvider::class)->addTag('nelmio_alice.faker.provider');
```
stof: you could wrap this in a `method_exists` check to keep support for the 2.8 LTS.

src/Bridge/Symfony/Resources/config/faker.xml
```diff
@@ -13,7 +13,7 @@
 
     <services>
         
-        <service id="Faker\Generator" public="false"
+        <service id="Faker\Generator" public="true"
```
stof: why public ?

PR #848, opened 2017-12-23
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertCount` instead of `count` function;
- `assertGreaterThanOrEqual` and `assertLessThanOrEqual` for mathematical comparion;
- `assertSame` instead of strict comparison `===`;
- `assertRegExp` instead of `preg_match`.

There is a `php_unit_dedicate_assert` rule for `PHP-CS-Fixer`. Should we enable it?

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -74,6 +74,7 @@ LICENSE
         'phpdoc_scalar' => true,
         'phpdoc_separation' => true,
         'phpdoc_trim' => true,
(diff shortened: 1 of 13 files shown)


theofidry:
>There is a php_unit_dedicate_assert rule for PHP-CS-Fixer. Should we enable it?

Yes let's do that 👍

PR #858, opened 2018-02-13
Refactoring FileReferenceResolver

### LINKED ISSUE #851: Alice calls methods under __calls twice, in certain cases
I'm not sure what exactly causes this behavior, but my observation is that if the following two requirements are met:

* You have a fixture A, that has a method listed under `__calls`
* Another fixture B has a `__construct`, where the argument is a reference to fixture A

Example:
```yaml
KN\Entity\Sporty:
    sporty1:
        __calls:
            - setOwningTeam: []

KN\Entity\SportyShareCode:
    code1:
        __construct: [ "@sporty1" ]

```

The method `setOwningTeam` will be called twice.

I have made a POC for this that you can clone easily:

https://github.com/norbert-yoimo/alice-poc-double-calls

If you run `php test.php`, you will see:

```
CALLED 1 times
CALLED 2 times
```

PHP version is 7.2.1, Alice version is 3.1.3.

Nijusan:
trying to fix #851 : adds setting CompleteObject for generated content in FileReferenceResolver; fixing tests.

the problem occured due to the following process:

DoublePassGenerator runs with firstPass over the fixtures, trying to generate objects with calling CompleteObjectGenerator calling SimpleObjectGenerator. 
So when this happens for 'code1', the instantiator->instantiate() will follow in calling the FixtureReferenceResolver cause auf the '_construct' - call in the fixture. This FixtureReference Resolves the reference to the fixture for sporty1 and will call the SimpleObjectGenerator->generate() for it.

The Problem here was, that this object was created, but the CompleteObjectGenerator does not track for it. This class only recognizes, that an IncompleteClass was created for 'code1' and correctly does not set this class as completed.
Now the DoublePassGenerator runs with secondPass and again, hydrates the fixture for sporty1 altough it was already hydrated to the FixtureReferenceResolver.

I saw in the FixtureReferenceResolverTest-Class, that it was assumed and expected, that the Generator will return a FixtureSet with an ObjectBag with an instance of CompleteObject. But this was only because the ObjectBag was initialised with it. In a non test environment the FixtureReferenceResolver would return a FixtureSet with an ObjectBag with SimpleObjects in it...
So i changed this behavior in the FixtureReferenceResolver to work as expected as in the Test-Class... The generated object will be set as CompleteObject and the Test-Class had to be changed a little.

### DIFF
--- a/src/Generator/Resolver/Value/Chainable/FixtureReferenceResolver.php
+++ b/src/Generator/Resolver/Value/Chainable/FixtureReferenceResolver.php
@@ -147,11 +147,15 @@ final class FixtureReferenceResolver implements ChainableValueResolverInterface,
 
             $context->markIsResolvingFixture($referredFixtureId);
             $objects = $this->generator->generate($referredFixture, $fixtureSet, $context);
-            $fixtureSet =  $fixtureSet->withObjects($objects);
 
             if (false === $needsCompleteGeneration) {
+                $generatedObject = $objects->get($referredFixture);
+                $objects = $objects->with(new CompleteObject($generatedObject));
+
                 $context->unmarkAsNeedsCompleteGeneration();
             }
+
+            $fixtureSet =  $fixtureSet->withObjects($objects);
 
             return new ResolvedValueWithFixtureSet(
                 $fixtureSet->getObjects()->get($referredFixture)->getInstance(),
--- a/tests/Loader/LoaderIntegrationTest.php
+++ b/tests/Loader/LoaderIntegrationTest.php
@@ -1871,6 +1871,36 @@ class LoaderIntegrationTest extends TestCase
             ]
         ];
 
+        yield '[construct] with reference to object with throwable setter and caller' => [
+            [
+                FixtureEntity\DummyWithThrowableSetter::class => [
+                    'another_dummy' => [
+                        '__calls' => [
+                            ['setVal' => [1]],
+                        ]
+                    ]
+                ],
+                FixtureEntity\DummyWithConstructorParam::class => [
+                    'dummy' => [
+                        '__construct' => [
+                            '@another_dummy'
+                        ]
+                    ]
+                ]
+            ],
+            [
+                'parameters' => [],
+                'objects' => [
+                    'another_dummy' => $anotherDummy1 = (function (FixtureEntity\DummyWithThrowableSetter $anotherDummy1) {
+                        $anotherDummy1->setVal(1);
+
+                        return $anotherDummy1;
+                    })(new FixtureEntity\DummyWithThrowableSetter()),
+                    'dummy' => $dummy1 = new FixtureEntity\DummyWithConstructorParam($anotherDummy1),
+                ]
+            ]
+        ];
+
         yield 'empty instance' => [
             [
                 stdClass::class => [
--- a/tests/Loader/LoaderIntegrationTest.php
+++ b/tests/Loader/LoaderIntegrationTest.php
@@ -1875,6 +1875,7 @@ class LoaderIntegrationTest extends TestCase
             [
                 FixtureEntity\DummyWithThrowableSetter::class => [
                     'another_dummy' => [
+                        'hydrated' => true,
                         '__calls' => [
                             ['setVal' => [1]],
                         ]
@@ -1893,7 +1894,7 @@ class LoaderIntegrationTest extends TestCase
                 'objects' => [
                     'another_dummy' => $anotherDummy1 = (function (FixtureEntity\DummyWithThrowableSetter $anotherDummy1) {
                         $anotherDummy1->setVal(1);
-
+                        $anotherDummy1->setHydrated(true);
                         return $anotherDummy1;
                     })(new FixtureEntity\DummyWithThrowableSetter()),
                     'dummy' => $dummy1 = new FixtureEntity\DummyWithConstructorParam($anotherDummy1),
--- a/tests/Loader/LoaderIntegrationTest.php
+++ b/tests/Loader/LoaderIntegrationTest.php
@@ -1861,12 +1861,12 @@ class LoaderIntegrationTest extends TestCase
             [
                 'parameters' => [],
                 'objects' => [
-                    'another_dummy' => $anotherDummy1 = StdClassFactory::create([
-                        'val' => 1,
-                    ]),
-                    'dummy' => $dummy1 = StdClassFactory::create([
-                        'val' => $anotherDummy1,
-                    ]),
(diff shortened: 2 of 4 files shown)


theofidry:
Thanks very much for looking into this @Nijusan. I'm aware this is a deep dive into the code and takes a lot of efforts. All of that work is very much appreciated.

Before I review this however (I'm not in a state to do so right now unfortunately), could you make sure you have a breaking test in `LoaderTestIntegration`?

The reason I'm a bit insisting on that one is because it's the closest you can get in this library to an e2e tests, which are by far the most reliable but in practice often ends up being horribly brittle and slow. Luckily enough, in this library this is not the case, so this test is quite very valuable and worth 10x different of unit tests.

Putted in another way: this is a kind of situation tricky enough for me to think that unit tests are not reliable enough to make sure we fixed the issue correctly and prevent in an efficient way any regression in the future.

Nijusan:
no problem: here we go https://github.com/nelmio/alice/pull/860 :)

src/Generator/Resolver/Value/Chainable/FixtureReferenceResolver.php
```diff
@@ -147,12 +147,16 @@ private function resolveReferredFixture(
 
             $context->markIsResolvingFixture($referredFixtureId);
             $objects = $this->generator->generate($referredFixture, $fixtureSet, $context);
-            $fixtureSet =  $fixtureSet->withObjects($objects);
 
             if (false === $needsCompleteGeneration) {
+                $generatedObject = $objects->get($referredFixture);
+                $objects = $objects->with(new CompleteObject($generatedObject));
```
theofidry: I'm a bit confused by this. IIRC the `$needsCompleteGeneration` flag is just to say "get this object, but fully instantiated". So your added code reads as "if the object doesn't need to be complete, mark it as complete" and that wouldn't make sense

tests/Loader/LoaderIntegrationTest.php
```diff
@@ -1861,12 +1861,12 @@ public function provideFixturesToGenerate()
             [
                 'parameters' => [],
                 'objects' => [
-                    'another_dummy' => $anotherDummy1 = StdClassFactory::create([
-                        'val' => 1,
-                    ]),
-                    'dummy' => $dummy1 = StdClassFactory::create([
-                        'val' => $anotherDummy1,
-                    ]),
+                    'another_dummy' => $anotherDummy1 = (function (FixtureEntity\DummyWithThrowableSetter $anotherDummy1) {
```
theofidry: _Maybe_ the reason why the tests still pass nonetheless are that I think this test is incomplete. The bug is that in this context, the object (`$anotherDummy`) is getting initialised + hydrated twice (instead of once — this is the bug) + getting the function calls.

I think the current tests don't account for the last call, so what I'm afraid is happening right now is initialised + hydrated (once — fix) ~~+ getting the function class~~ (and this is not good)

So if my assumption is correct, if we add a function call, e.g. `DummyWithThrowableSetter::complete()` to `$anoherDummy1` the tests would fail again with your current patch.

Provided the above is correct, we need to find something else to fix this. I can't take a look at the moment, but the issue is that `$anotherDummy1` is required at some point, it is getting initialised and everything, but for some reason even though it has been created as a complete object it has not been marked as such.

src/Generator/Resolver/Value/Chainable/FixtureReferenceResolver.php
Nijusan: hmm, for me it reads like "the object does not need to be completed furthermore, so first, mark it as completed and second unmark the needs completion flag ... because its already completed...

the other way around, if i would put that code elsewhere, it would read like "the object is still marked as "needs completion" but i would already mark it as completed, but why? 

i am not 100% sure about this flag... all i could find out from the source until now the flag seems to be only used in this very special constellation, to make it possible to hydrate the object although we are in the first pass... and if thats the point, then saving the "needsCompleteGeneration" flag and setting it back later would make no use at all... because what we want in this special situation is to have the object to be fully complete?!

tests/Loader/LoaderIntegrationTest.php
Nijusan: sry, i've lost u ... 

i've nothing changed on the function calls? they should still be called? or am i wrong? from the point of the original https://github.com/nelmio/alice/issues/851 issue, i think the problem does not occur because of the use of __call but because of the use of the __construct .

i'm not sure about ur point i think i did not understand it correctly. nonetheless i added another test case, maybe this is going the direction ur talking about?

theofidry: Ok let me rephrase it.

So we have two things here: the objects and the fixtures. The fixture describes what the object looks like (how it's constructed, initialised etc.). So what alice do is:

- go through each fixture
- instantiate the object for each fixture (this is the first pass in the objects generator)
- go through each fixture again an "complete" the objects which involves two steps:
    - hydrating the object
    - configuring the object (doing the calls registered in `__call`)
    - mark the object as complete

So an object is complete only when it went through instantiated + hydrated + configured.

However things are not so simple, because an object may need another. For example here `$dummy` needs `$anotherDummy1`. When this happens, we try to get `$anotherDummy1` as a _complete_ object. At this point of time `$anotherDummy1` may not be created or created but not completed yet. So we need to complete `$anotherDummy1` first before passing it to `$dummy`.

I think the bug here is that when `$dummy` needs `$anotherDummy1`, we do complete it properly and then pass it to `$dummy`. The bug however is that when we do that, we fail to mark `$anotherDummy1` as a complete object (i.e. make it an instance of `CompleteObject` instead of `SimpleObject`). So later during the second pass, `$anotherDummy1` is being completed again which is the bug

Nijusan: ok. but what u are writing is what i was trying to say right from the beginning :) after diggin deep into the code, exactly what u say is what i could reproduce... 

and because of this i do make the object an instance of CompleteObject so that it is not completed again in second pass because everthing of this happend in the first pass...

so i think we are both talking same language here? what is it that makes u unsatisfied ?

theofidry: Oh I'm not unsatisfied at all :) I just wanna make sure we are actually fixing this and not causing another issue at the same time.

So what I think is missing in the test is a `__call` on `$anotherDummy1` like you have for hydrating its value. This is just to make sure that `$antoherDummy1` is properly passed as complete (it gets both hydrated & configured) and that no second call is made for it. To rephrase: you added a test accounting for the hydration, I would add a second for the "config" part.

theofidry:
And here we go! Thanks very much. I'll do a release ASAP

ogizanagi:
✨  Thanks @Nijusan for having taken care of this :)

PR #866, opened 2018-02-20
adds check for service tag

Nijusan:
currently alice and aliceBundle both try to autoRegister classes which are extending BaseFakerProvider.

preventing this happening until alice bundle bumps version information about require nelmio/alice

### DIFF
--- a/src/Bridge/Symfony/DependencyInjection/NelmioAliceExtension.php
+++ b/src/Bridge/Symfony/DependencyInjection/NelmioAliceExtension.php
@@ -70,6 +70,8 @@ final class NelmioAliceExtension extends Extension
             );
         }
 
+        // Check if the auto-configuration of the tag is not already done. This is a temporary measure to avoid to break
+        // existing versions of HautelookAliceBundle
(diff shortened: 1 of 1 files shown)


src/Bridge/Symfony/DependencyInjection/NelmioAliceExtension.php
```diff
@@ -70,6 +70,8 @@ private function loadServices(ContainerBuilder $container)
             );
         }
 
-        $container->registerForAutoconfiguration(BaseFakerProvider::class)->addTag('nelmio_alice.faker.provider');
+        if ( 0 === count($container->findTaggedServiceIds('nelmio_alice.faker.provider')) ) {
```
theofidry: Could you add the following comment? just above it?

```
// Check if the auto-configuration of the tag is not already done. This is a temporary measure to avoid to break
// existing versions of HautelookAliceBundle
```

theofidry:
Looks good except a minor comment :)

theofidry:
Thanks!

PR #873, opened 2018-02-23
Fix the readme

stof:
- Fix the link to the BC promise
- Fix the Symfony Slack invitation link. Slack invitation links have a limited lifetime (they expire after a fixed number of usages). So symfony.com has its own URL which redirects to the invitation link and is updated when they generate a new link.


theofidry:
Thanks! I'll probably need to check the other projects as well regarding the link, I added it before we had this symfony.com URL

PR #880, opened 2018-02-27
Fixed minimum parameter count.

ChristianVermeulen:
screenshot 2018-02-27 17 57 26

There is a wrong conditional for this case :-D.


theofidry:
Indeed! Thanks :)

PR #883, opened 2018-03-03
Do not deep clone resolved values

### LINKED ISSUE #859: Add relation to same instance in differebt entities
Hi there, 

I have such  entities structure:
User has many Campaigns
User has many Creatives
Campaign has many Creatives

And  .yml file:
```
Project\Entities\User:
  user{1..5}:
    email (unique): '<email()>'
    firstName: '<firstName()>'
    lastName: '<lastName()>'
    password: '<bcrypt("q1w2e3r4")>'

Project\Entities\Campaign:
  campaign{1..100}:
    user: '@user*'
    name: '<company()>'

Project\Entities\Creative:
  creative{1..1000}:
    name: '<word()>'
    campaign: '@campaign*'
    user: '@user*' //i<---------issue is here
```
So  issue is in Creative part, I want to set user value same as we use in a related campaign (@campaign*), but not sure how to do that.

I tried:
```
Project\Entities\Creative:
  creative{1..1000}:
    name: '<word()>'
    campaign: '@campaign*'
    user: '$campaign->user'
```

but got the error   **Could not lex the value ">user"** , also I tried

```
Project\Entities\Creative:
  creative{1..1000}:
    name: '<word()>'
    campaign: '@campaign*'
    user: '@self->campaign->user'
```

and got the error   **Could not parse the token "@self->campaign->user" (type: PROPERTY_REFERENCE_TYPE).** 

Any ideas how to solve it?

ixarlie:
FunctionCallArgumentResolver when resolving every argument in the FunctionCallValue, creates a new FunctionCallValue to be resolved by any of the functions callers.

FunctionCallValue creates a clone for every argument, so this could be a problem if we want to use objects as arguments (references, variables, ...). For example: to relate two objects that should be persisted. The cloned object is totally new for the persister and out of scope.

Fixes #859

### DIFF
--- a/src/Definition/Value/ResolvedFunctionCallValue.php
+++ b/src/Definition/Value/ResolvedFunctionCallValue.php
@@ -16,7 +16,7 @@ namespace Nelmio\Alice\Definition\Value;
 use Nelmio\Alice\Definition\ValueInterface;
 
 /**
- * Value object representing '<name()>' (no arguments cloning)
+ * Value object representing a function arguments which have already been re-solved.
  */
 final class ResolvedFunctionCallValue implements ValueInterface
 {
--- a/src/Definition/Value/ResolvedFunctionCallValue.php
+++ b/src/Definition/Value/ResolvedFunctionCallValue.php
@@ -1,0 +1,71 @@
+<?php
+
+/*
+ * This file is part of the Alice package.
+ *
+ * (c) Nelmio <hello@nelm.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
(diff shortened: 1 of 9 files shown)


src/Definition/Value/ResolvedFunctionCallValue.php
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Nelmio\Alice\Definition\Value;
+
+use Nelmio\Alice\Definition\ValueInterface;
+
+/**
+ * Value object representing '<name()>' (no arguments cloning)
```
theofidry: representing a function arguments which have already been re-solved.

ixarlie:
Thanks for your comments!

theofidry:
Thanks very much, nice work :)

PR #885, opened 2018-03-04
Symfony Integration Doc - Current provider

### LINKED ISSUE #884: Symfony Bridge auto add current faker method to blacklist if is not included.
It is a really minor issue but if `<current()>` is an internal alice faker, when I edit blacklisted functions like this, `current` stop working because is trying to use PHP function instead.
```yml
nelmio_alice:
# ...
  functions_blacklist:
    - 'md5'
# ...
```

Ok, the solution is simple, add `current` to the list but I think it should be added by alice if it is not in the list.

ixarlie:
resolves #884 

Any improve in the text is appreciated.

### DIFF
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -166,6 +166,8 @@ nelmio_alice:
                                    # failing.
 ```
 
+Note: When using `<current()>` with the Alice built-in provider, be sure `current`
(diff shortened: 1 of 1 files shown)


theofidry:
Thanks!

Issue #888, opened 2018-03-07
Using (unqiue) on referenced values results in duplicated entries.


vgeyer:
Hi, 
i found a very unexpected behaviour using (unique) on relation field:
```yaml
App\Entity\Payment\PaymentProvider:
  payment_provider{1..2}:
    shortName: "<current()>"
    longName: "<name()>"
App\Entity\Payment\PaymentInterface:
  payment_interface{1..2}:
    paymentProvider (unique): "@payment_provider*"
    application: "production"
```
I already explained it a little bit more detailed [here](https://github.com/nelmio/alice/issues/682#issuecomment-370930709) but back then i wasnt paying attention to the (unique). I located the error to be somewhere in ```\Nelmio\Alice\Generator\Resolver\Value\Chainable\UniqueValueResolver``` or its resolvers. 
The expected behaviour would use an existing instance, not allready picked instance, or throw an error when you got something like this:
```yaml
App\Entity\Payment\PaymentProvider:
  payment_provider1:
    shortName: "1"
    longName: "1"
App\Entity\Payment\PaymentInterface:
  payment_interface{1..2}:
    paymentProvider (unique): "@payment_provider*"
    application: "production"
```

I "fixed" my problem ensuring the uniqueness by working with current and this somehow fixes the error. So either we should put a warning on the documentation for (unique) or fix this. Personaly i will stop here, because i allready ivested a huge amount of time in this... maybe in me free time tho

theofidry:
Hi, as you can see in #899 I cannot reproduce this behaviour. So I guess there is more to it?

theofidry:
Closing as was not able to reproduce it with #888. Please feel free to open a new issue if you still encounter the bug and manage to get a reproducer

PR #899, opened 2018-03-17
Add test case for #888

theofidry:


theofidry:
I think it served its purpose so closing now

PR #903, opened 2018-03-21
Add range name reference denormalizer

reinfi:
Hey,

I tried to port my reference range name builder to version 3.0. It works but I'm quite unsure if this is the right way to handle it. (Solves https://github.com/nelmio/alice/issues/692)

One point which is currently not working is the <current()> value, because the simple fixture just requires a string as a name. That would work but then I would need to translate that to the referenced fixture by the id. The ValueForCurrentValueResolver would be the right place but I'm unsure how to differentiate between a normal current value and a current value from the referenced range.

I could add my own Fixture class but just for the one denormalizer?

As I don't know if that is the right way to implement it, I have not add any tests, would do this later when you say that it is the right way to handle the requirements.

And I am a bit unhappy about the name, so if you have any suggestions :)

Thanks for your help.

### DIFF
--- a/doc/complete-reference.md
+++ b/doc/complete-reference.md
@@ -130,21 +130,24 @@ You can also specify a reference to a previously created list of fixtures:
 Nelmio\Entity\User:
     user_{1..10}:
         username: '<name()>'
+
 Nelmio\Entity\UserDetail:
     userdetail_{@user_*}:  # is going to generate `userdetail_user_1`, `userdetail_user_2`, ..., `userdetail_user_10`
         user: <current()>
         email: '<email()>'
 ```
+
 You could either use a star to get all created fixtures matched by the reference or use just one by giving the full fixture name.
 
 ```yaml
 Nelmio\Entity\User:
     user_bob:
         username: 'bob'
+
 Nelmio\Entity\UserDetail:
     userdetail_{@user_bob}:
-        user: <current()>
-        email: 'bob@test.de'
+        user: <current()>   # holds `@user_bob`
+        email: 'bob@test.de'
 ```
 
 >The `<current()>` function holds the value of the referenced fixture.
--- a/doc/complete-reference.md
+++ b/doc/complete-reference.md
@@ -4,6 +4,7 @@
     1. [YAML](#yaml)
     1. [PHP](#php)
 1. [Fixture Ranges](#fixture-ranges)
+1. [Fixture Reference](#fixture-reference)
 1. [Fixture Lists](#fixture-lists)
 1. [Calling Methods](#calling-methods)
     1. [Method arguments with flags](#method-arguments-with-flags)
@@ -130,7 +131,7 @@ Nelmio\Entity\User:
     user_{1..10}:
         username: '<name()>'
 Nelmio\Entity\UserDetail:
-    userdetail_{@user_*}:
+    userdetail_{@user_*}:  # is going to generate `userdetail_user_1`, `userdetail_user_2`, ..., `userdetail_user_10`
         user: <current()>
         email: '<email()>'
 ```
--- a/doc/complete-reference.md
+++ b/doc/complete-reference.md
@@ -121,6 +121,32 @@ Nelmio\Entity\User:
 
 To go further we the example above, we can just randomize data.
 
+## Fixture Reference
+
+You can also specify a reference to a previously created list of fixtures:
+
+```yaml
+Nelmio\Entity\User:
+    user_{1..10}:
+        username: '<name()>'
+Nelmio\Entity\UserDetail:
+    userdetail_{@user_*}:
+        user: <current()>
+        email: '<email()>'
+```
+You could either use a star to get all created fixtures matched by the reference or use just one by giving the full fixture name.
+
+```yaml
+Nelmio\Entity\User:
+    user_bob:
+        username: 'bob'
+Nelmio\Entity\UserDetail:
+    userdetail_{@user_bob}:
+        user: <current()>
+        email: 'bob@test.de'
+```
+
+>The `<current()>` function holds the value of the referenced fixture.
 
 ## Calling Methods
 
--- a/fixtures/User.php
+++ b/fixtures/User.php
@@ -15,10 +15,21 @@ namespace Nelmio\Alice;
 
 class User
 {
-    /**
-     * @var string|null
-     */
+    private $id;
+
     private $name;
+
+    public function getId(): string
+    {
+        return $this->id;
+    }
+
+    public function setId(string $id): User
+    {
+        $this->id = $id;
+
+        return $this;
+    }
 
     public function getName(): ?string
     {
--- a/fixtures/User.php
+++ b/fixtures/User.php
@@ -1,33 +1,34 @@
 <?php
 
+/*
+ * This file is part of the Alice package.
+ *
+ * (c) Nelmio <hello@nelm.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
 namespace Nelmio\Alice;
 
-/**
- * @package Nelmio\Alice
- */
 class User
 {
+    /**
+     * @var string|null
+     */
     private $name;
 
-    /**
-     * @return mixed
-     */
-    public function getName()
+    public function getName(): ?string
     {
         return $this->name;
     }
 
-    /**
-     * @param mixed $name
-     *
-     * @return User
-     */
-    public function setName($name)
+    public function setName(?string $name): User
     {
         $this->name = $name;
 
         return $this;
     }
-
-
 }
--- a/fixtures/User.php
+++ b/fixtures/User.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Nelmio\Alice;
+
+/**
+ * @package Nelmio\Alice
+ */
+class User
+{
+    private $name;
+
+    /**
+     * @return mixed
+     */
+    public function getName()
+    {
+        return $this->name;
+    }
+
+    /**
+     * @param mixed $name
+     *
+     * @return User
+     */
+    public function setName($name)
+    {
+        $this->name = $name;
+
+        return $this;
+    }
+
+
+}
--- a/fixtures/UserDetail.php
+++ b/fixtures/UserDetail.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Nelmio\Alice;
+
+/**
+ * @package Nelmio\Alice
+ */
+class UserDetail
+{
+    private $email;
+
+    /**
+     * @var User
+     */
+    private $user;
+
+    /**
+     * @return mixed
+     */
+    public function getEmail()
+    {
+        return $this->email;
+    }
+
+    /**
+     * @param mixed $email
+     *
+     * @return UserDetail
+     */
+    public function setEmail($email)
+    {
+        $this->email = $email;
+
+        return $this;
+    }
+
+    /**
+     * @return User
+     */
+    public function getUser(): User
+    {
+        return $this->user;
+    }
+
+    /**
+     * @param User $user
+     *
+     * @return UserDetail
+     */
+    public function setUser(User $user)
+    {
+        $this->user = $user;
+
+        return $this;
+    }
+}
--- a/fixtures/UserDetail.php
+++ b/fixtures/UserDetail.php
@@ -1,12 +1,23 @@
 <?php
 
+/*
+ * This file is part of the Alice package.
+ *
+ * (c) Nelmio <hello@nelm.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
 namespace Nelmio\Alice;
 
-/**
- * @package Nelmio\Alice
- */
 class UserDetail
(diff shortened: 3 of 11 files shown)


theofidry:
Hi @reinfi, thanks for porting that to 3.x!

To to be clear on what we want to do, with the following example:

```yaml
Nelmio\Entity\UserDetail:
    # Reference by concrete id
    userdetail{@user1}:
        user: <current()>
        username: <username()>
        fullname: <firstname()> <lastname()>
        birthDate: 1980-10-10
        email: <current()>@example.org
        favoriteNumber: 42
    # Reference by a list of fixtures    
    userdetail{@user*}:
        user: <current()>
        username: <username()>
        fullname: <firstname()> <lastname()>
        birthDate: 1980-10-10
        email: <current()>@example.org
        favoriteNumber: 42
```

The result will be:

- user1 (users are defined elsewhere)
- ...
- userN
- userdetail_user1 -> actually defined 2 times but only the last definition will remain
- ...
- userdetail_userN

So the first step is the normalization: the normalization is meant to be a where we transform the data input into a set of comprehensible objects. So here we'll need somehow to interpret `userdetail{@user*}` in a way so that we can resolve it later.

Further in the process, before starting to create the objects, we resolve the `FixtureSet` where we interpret the templates and co. It is at this step that we will need to resolve `userdetail{@user*}` into `userdetail_user1`, ..., `userdetail_userN`.
Because of its dynamic nature, I think this needs to be done it two passes. The first pass would be what is currently done, so the lists like  `user{1..10}` will be transformed into `user1`, ..., `user10`and so on. Only the dynamic range fixtures like `userdetail{@user*}` will remain.
We can then initiate the second pass where we resolve those `userdetail{@user*}` to transform it into `userdetail_user1`, ..., `userdetail_userN`.

And I actually don't think we will need anything more. If you don't want to go too far with the tests yet, I recommend to try to implement an integration test in `LoaderIntegrationTest`. A single test to try to make this feature work, and then when this is working you can add more tests and cleanup

reinfi:
Thanks for you answers as long as I have not understand it in the context of my questions.

Good think, we are talking about the same issue :-)

So from your answer I think it is the right way to implement it.

Still the question how to deal with the `<current()>` value problem. I could add a flag but this would not the same as currently flags are defined and used.

I have already a test in my branch to test the feature so, no problem about writing it :)

theofidry:
I think we can go through the fixture values to replace the occurrences of `<current()>` by the reference value

reinfi:
So I finished my work at the moment. If you have any suggestions feel free to share :)

reinfi:
@theofidry  any idea when we can continue with this? Because this blocks our project as we can't migrate to alice 3.x what I would really like to do.

Hope this does not last as long as the pr for 2.x :)

theofidry:
Sorry I missed your previous comment. I'll try to review this before the end of the week!

fixtures/User.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Nelmio\Alice;
+
+/**
+ * @package Nelmio\Alice
```
theofidry: let's remove that phpdoc, unneeded

fixtures/User.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+class User
+{
+    private $name;
```
theofidry: since `$name` is not set in a constructor I think it's a good practice to add a phpdoc so here `@var string|null`

fixtures/User.php
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * @return mixed
+     */
+    public function getName()
```
theofidry: we're in PHP 7.1 so we can add the return type `?string` and remove the redundant/useless phpdoc

fixtures/User.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        return $this->name;
+    }
+
+    /**
+     * @param mixed $name
+     *
+     * @return User
+     */
+    public function setName($name)
```
theofidry: same as the comment above regarding the strict types & the phpdoc

fixtures/User.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        $this->name = $name;
+
+        return $this;
+    }
+
+
+}
```
theofidry: missing ending line return and there is two extra blank lines above as well

fixtures/UserDetail.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace Nelmio\Alice;
+
+/**
+ * @package Nelmio\Alice
+ */
+class UserDetail
```
theofidry: same comments as for the previous class regarding the CS, phpdoc & strict types

src/Definition/Fixture/SimpleFixture.php
```diff
@@ -38,11 +38,11 @@
     private $specs;
 
     /**
-     * @var string
+     * @var string|int|FixtureInterface
      */
     private $valueForCurrent;
 
-    public function __construct(string $id, string $className, SpecificationBag $specs, string $valueForCurrent = null)
+    public function __construct(string $id, string $className, SpecificationBag $specs, $valueForCurrent = null)
```
theofidry: I would add the phpdoc for the param `$valueForCurrent` as well now then (just that param) besides in the var L41 like you did. The reason being an IDE will be able to check for the type during auto-completion

src/FixtureBuilder/Denormalizer/Fixture/Chainable/NullRangeNameReferenceDenormalizer.php
```diff
@@ -0,0 +1,193 @@
...
+use Nelmio\Alice\FixtureBuilder\Denormalizer\Fixture\ChainableFixtureDenormalizerInterface;
+use Nelmio\Alice\FixtureBuilder\Denormalizer\Fixture\SpecificationsDenormalizerInterface;
+use Nelmio\Alice\FixtureBuilder\Denormalizer\FlagParserAwareInterface;
+use Nelmio\Alice\FixtureBuilder\Denormalizer\FlagParserInterface;
+use Nelmio\Alice\FixtureInterface;
+use Nelmio\Alice\IsAServiceTrait;
+use Nelmio\Alice\Throwable\Exception\LogicExceptionFactory;
+
+final class NullRangeNameReferenceDenormalizer
+    implements ChainableFixtureDenormalizerInterface, FlagParserAwareInterface
```
theofidry: I'm not fond of the line return for the `implements` even if that makes the line extra long it's fine

src/FixtureBuilder/Denormalizer/Fixture/Chainable/NullRangeNameReferenceDenormalizer.php
```diff
@@ -0,0 +1,193 @@
...
+{
+    use IsAServiceTrait;
+
+    /** @private */
+    const REGEX = '/.+\{(?<expression>@(?<name>([A-Za-z0-9-_]+))(?<flag>(\*+))?)\}/';
```
theofidry: we're in 7.1 so I believe we can do `private const` now!

src/FixtureBuilder/Denormalizer/Fixture/Chainable/NullRangeNameReferenceDenormalizer.php
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     * @inheritdoc
+     */
+    public function denormalize(
+        FixtureBag $builtFixtures,
+        string $className,
+        string $fixtureId,
+        array $specs,
+        FlagBag $flags
+    ): FixtureBag {
```
theofidry: in that peculiar case:

```diff
-): FixtureBag {
+): FixtureBag
+{
```

theofidry: same for the other methods

src/FixtureBuilder/Denormalizer/Fixture/Chainable/NullRangeNameReferenceDenormalizer.php
```diff
@@ -0,0 +1,193 @@
...
+                    $valueForCurrent
+                )
+            );
+        }
+
+        return $builtFixtures;
+    }
+
+    /**
+     * @param FixtureBag $builtFixtures
```
theofidry: let's get rid of the phpdoc if it doesn't provide any additional value to the typehint :)

src/FixtureBuilder/Denormalizer/Fixture/Chainable/NullRangeNameReferenceDenormalizer.php
```diff
@@ -0,0 +1,193 @@
...
+        return $matchedFixtures;
+    }
+
+    /**
+     * @param string $fixtureId
+     *
+     * @return string
+     * @throws \LogicException
+     */
+    private function determineFixtureIdPrefix(
```
theofidry: it feels like this could be on one line

src/FixtureBuilder/Denormalizer/Fixture/Chainable/NullRangeNameReferenceDenormalizer.php
```diff
@@ -0,0 +1,193 @@
...
+        );
+
+        return $matchedFixtures;
+    }
+
+    /**
+     * @param string $fixtureId
+     *
+     * @return string
+     * @throws \LogicException
```
theofidry: in general it's unnecessary to document `\LogicException`s: they are sign of an incorrect config requiring manual fix, i.e. it's not an error you should ever encounter at runtime (unlike `\RuntimeException`) and account for

src/FixtureBuilder/Denormalizer/Fixture/Chainable/NullRangeNameReferenceDenormalizer.php
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * @param string           $fixtureId
+     * @param string           $className
+     * @param array            $specs
+     * @param FlagBag          $flags
+     * @param FixtureInterface $valueForCurrent
+     *
+     * @return FixtureInterface
+     * @throws \Nelmio\Alice\Throwable\DenormalizationThrowable
```
theofidry: this class can be imported; since the other phpdoc elements are redundant we can get rid of the others

tests/Loader/LoaderIntegrationTest.php
```diff
@@ -581,6 +583,43 @@ public function testLoadSelfReferencedFixtures()
...
+        $set = $this->loader->loadData($data);
+
+        $objects = $set->getObjects();
+        $this->assertCount(5, $objects);
+
+        $this->assertArrayHasKey('userdetail_user0', $objects);
+        $this->assertArrayHasKey('userdetail_user1', $objects);
+        $this->assertArrayHasKey('userdetail_single_user1', $objects);
+
+        $this->assertInstanceOf(User::class, $objects['userdetail_user0']->getUser());
```
theofidry: maybe more than `assertInstanceOf()` we could do a `assertSame()` as the instance should be the same

tests/Loader/LoaderIntegrationTest.php
```diff
@@ -581,6 +583,43 @@ public function testLoadSelfReferencedFixtures()
         $this->assertCount(2, $objects);
     }
 
+    public function testLoadRangeReference()
+    {
+        $data = [
+            User::class => [
+                'user0' => [
```
theofidry: what happens if we have a template user, e.g. `base_user (template)` and the `user0` would extend it? (I wouldn't expect the `base_user` to be picked up)

reinfi: Good point, skipped now template fixtures when building the list.

src/FixtureBuilder/Denormalizer/Fixture/Chainable/NullRangeNameReferenceDenormalizer.php
reinfi: That was the ******** phpstorm bug with exceptions :-(

doc/complete-reference.md
```diff
@@ -121,6 +121,32 @@ Nelmio\Entity\User:
...
 
+## Fixture Reference
+
+You can also specify a reference to a previously created list of fixtures:
+
+```yaml
+Nelmio\Entity\User:
+    user_{1..10}:
+        username: '<name()>'
+Nelmio\Entity\UserDetail:
```
theofidry:
```yaml
Nelmio\Entity\User:
    user_{1..10}:
        username: '<name()>'

Nelmio\Entity\UserDetail:
    userdetail_{@user_*}: # is going to generate `userdetail_user_1`, `userdetail_user_2`, ..., `userdetail_user_10`
        user: <current()>
        email: '<email()>'
```

doc/complete-reference.md
```diff
@@ -121,6 +121,32 @@ Nelmio\Entity\User:
 
 To go further we the example above, we can just randomize data.
 
+## Fixture Reference
```
theofidry: you also need to add it to the ToC of this file besides the readme

reinfi:
Builds green (check)
Docs hopefully completed (check)

theofidry:
Thanks very much @reinfi, great work :)

PR #905, opened 2018-03-24
Allow one item collections

### LINKED ISSUE #891: List requires more than one item
There is a use case that I think has gone under the radar where you may define a fixture that looks something like the following:

```YAML
MyClass:
  person_{alice}:
    name: '<current()>'
```

At first this looks stupid (and I can agree) however this is a common case when extending from a template that defines the usage of `<current()>`. For example we have a template that is used to define groups of something, to make this easier it uses the current marker but in the cases where a group only contains one we are left with `{single}`.

I am curious if this is was done on purpose?

On another note I see there is another issue where spaces are required too now? #872

theofidry:
Closes #891

/cc @reinfi


reinfi:
Okay that was my thought too.

PR #906, opened 2018-03-26
Fix faker regexp when we want to generate SR_LATN names

tifabien:
Fix the regex when we want to generate SR_LATN names.

### DIFF
--- a/src/Nelmio/Alice/Instances/Processor/Methods/Faker.php
+++ b/src/Nelmio/Alice/Instances/Processor/Methods/Faker.php
@@ -120,7 +120,7 @@ class Faker implements MethodInterface
      */
     public function process(ProcessableInterface $processable, array $variables)
     {
-        $fakerRegex = '<(?:(?<locale>[a-z]+(?:_[a-z]+)?):)?(?<name>[a-z0-9_]+?)?\((?<args>(?:[^)]*|\)(?!>))*)\)>';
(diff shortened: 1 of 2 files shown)


theofidry:
Hi! Thanks for you contribution. I'm perfectly fine with the change but it would be neat if you could add a test for it as well. I think a simple case in `LoaderIntegrationTest` in `provideFixturesToGenerate` or something like that would do :)

tifabien:
The LoaderIntegrationTest class seems to be only on the master branch. Is there any equivalent for the 2.x branch?

theofidry:
Sorry I didn't notice this was for 2.x. You have [LoaderTest.php](https://github.com/nelmio/alice/blob/2.x/tests/Nelmio/Alice/Fixtures/LoaderTest.php)

tifabien:
I just added the unit test. I'll have a look at the 3.x branch and let you know if it's OK!

tifabien:
Just checked and all seems OK on the master branch :)

theofidry:
Alright, I'll try to have a look this week. I'm ok with the feature but I would like the build to be green first :)

tifabien:
it seems related to the deprecation notice introduced in SF 3.4 regarding the tests/Nelmio/Alice/Fixtures/Parser/Files/Yaml/file_with_constants.yml:10 file

theofidry:
Yep, it's probably easy to fix, but it needs to be fixed first nonetheless

theofidry:
Thank you @tifabien!

Issue #539, opened 2016-09-06
Add support for steps with ranges


aledeg:
It could be neat to use the range with steps. For instance, to create 5 different users:

``` yaml
User:
  User{1..10..2}:
    ...
```

It could be neat to use the range in reverse. For instance:

``` yaml
User:
  User{10..1}:
    ...
```

Those feature would be really handy with templates. The first one to get cyclic data. The second one to save current values in reverse.

theofidry:
Regarding the first one I find the syntax a bit confusing, maybe:

```yaml
User:
  User{1..10, 2}:
    #...
```

As for the later one this syntax is already allowed, but as the order does not matter it makes no difference to have `User1` first or `User10`.

aledeg:
>Regarding the first one I find the syntax a bit confusing

This is the [bash increment brace expansion feature syntax](http://wiki.bash-hackers.org/syntax/expansion/brace#increment). But I am ok with any syntax.
If I can find some time, I'll try to send a PR.

>As for the later one this syntax is already allowed, but as the order does not matter it makes no difference to have User1 first or User10.

I agree that on the data level it makes no differences but on the database level it makes a difference. The natural order of the database will send me the first inserted record. In my example, I was hoping that the first record would User10, but it was User1.

I made the following workaround for my test case:
```yaml
User:
    User{10}:
        ...
    User{1}:
        ...
```
This way, User10 is inserted before User9 which is inserted before User8 which is... but you get the idea.
With the braces around the number, I still can use ```<current()>``` in my object fields.

theofidry:
>I agree that on the data level it makes no differences but on the database level it makes a difference. The natural order of the database will send me the first inserted record. In my example, I was hoping that the first record would User10, but it was User1.

I'm not sure to understand the problem, if the order is important, then it should be re-ordered. It is extremely difficult to guarantee the order and there is several issues:

- how do you deal with injected objects, should they come first or after?
- the generation order is not the same as the declaration order, some objects requires other objects to be instantiated first
- there is several extensibility points, e.g. to add custom flags, where you can alter the order of fixtures

A solution is to order all the fixtures according to a certain criteria, but as this criteria is likely to be user specific and has a performance impact, I would rather not try to solve this in alice.

theofidry:
Edit regarding my last message:

- I'm cool with [this syntax](https://github.com/nelmio/alice/issues/539#issuecomment-244927575)
[...]

PR #917, opened 2018-04-07
Add support for steps with ranges (#539)

reinfi:
Resolves issue https://github.com/nelmio/alice/issues/539 for the step part.

Do I need to add the same functionality to FixtureRangeReferenceTokenParser?

### DIFF
--- a/doc/complete-reference.md
+++ b/doc/complete-reference.md
@@ -93,6 +93,19 @@ Now it will generate ten users, with IDs `user1` to `user10`. Pretty good but
 we only have 10 bobs with the same name, username and email, which is not
 so fancy yet.
 
+If you need gaps between the numbers you can add a step argument.
+
+```yaml
+Nelmio\Entity\User:
+    user{1..10, 2}:
+        username: bob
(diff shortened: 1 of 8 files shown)


theofidry:
>Do I need to add the same functionality to FixtureRangeReferenceTokenParser?

Could you explain in more details what you mean? Sorry I'm a bit out of the context :)

Regarding the PR all green on my side for now :)

reinfi:
The FixtureRangeReferenceTokenParser has the same logic for the range as the Denormalizer, but it is not used when I debug the loader test so I'm unsure if it also requires the step logic.

theofidry:
Hm maybe it's missing out a test in the `LexerIntegrationTest` and `ParserIntegrationTest`. It's weird that this parser is not called at all

reinfi:
That's what I did now. But I thought the parser should be called because it is designed to handle the user_{1..3} case. Now everything seems good, all tests are running :)

theofidry:
All good then, thanks very much

Issue #916, opened 2018-04-06
settype documentation issue?


ostrolucky:
[docs](https://github.com/nelmio/alice/blob/master/doc/customizing-data-generation.md#cast) recommend to cast, we should use `settype`. Here are the problems:

- Example lists arguments of this function reversed
- It returns boolean
- It requires value to be passed by reference. I think you solved this issue somehow on your side, but it's confusing anyway.

Instead, user should be directed to functions like `intval()` etc

theofidry:
It's true with have those casting functions so we should go with that instead indeed 👍

ostrolucky:
Thank you @reinfi

PR #918, opened 2018-04-08
settype documentation issue

reinfi:
Documentation fix for https://github.com/nelmio/alice/issues/916

I also ran the codestyle fixer to fix several classes.


theofidry:
I'm surprised some of those are not picked by the CS fixer...

Thanks very much :)

Issue #825, opened 2017-10-25
Add support load fixture from JSON-files


ArmRes:
Add support load fixture from JSON-files.

theofidry:
There is already an example for the YAML and PHP so this shouldn't be too complicated. PRs are welcomed :)

PR #919, opened 2018-04-08
Add JSON Parser

reinfi:
Solves https://github.com/nelmio/alice/issues/825

Is the JsonParserTest enough?


reinfi:
how silly :-D 

Now it is registered and I added test to the LoaderIntegrationTest.

theofidry:
Thanks very much @reinfi :)

PR #926, opened 2018-05-24
Fix property reference usage with a function call

### LINKED ISSUE #925: Property reference combined with interpolated relation (<x>) fails with exception
Here is the use case:
```yaml
Subject1:
    subject1_instace_{1..100}:
        prop: '<whatever>'
        
Subject2:
    subject2_instace_{1..100}:
        prop: '@subject1_instace_<current()>->prop'
```

results in: 
```
 [Nelmio\Alice\Throwable\Exception\FixtureBuilder\ExpressionLanguage\LexException]  
  Could not lex the value ">prop"
```

Using only the relation (`@subject1_instace_<current()>`) or only the property reference (`@subject1_instace_1->prop`) works as expected. 

Tested on v3.2.2 and v3.3.0

ssaki:
- Added a lexer regex to match a property reference using a func call
- relaxed the check of the function tokenizer as it will miss the already tokenized function of the reference and fail during the second pass (invoked by PropertyReferenceTokenParser)

Closes #925

### DIFF
--- a/src/FixtureBuilder/ExpressionLanguage/Lexer/FunctionTokenizer.php
+++ b/src/FixtureBuilder/ExpressionLanguage/Lexer/FunctionTokenizer.php
@@ -57,7 +57,7 @@ final class FunctionTokenizer
 
     public function isTokenized(string $value): bool
     {
-        return '<aliceTokenizedFunction(' === substr($value, 0, 24);
+        return strpos($value, '<aliceTokenizedFunction(') !== false;
     }
 
     public function detokenize(string $value): string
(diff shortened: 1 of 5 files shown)


theofidry:
Thanks for looking it up 👍 ! Looks good to me except I would also add a test to the `LexerIntegrationTest` and `ParserIntegrationTest` as the unit test by itself doesn't guarantee much (the whole expression language behaviour changes a lot depending on how you configure it...)

ssaki:
@theofidry , I've added test cases in both `LexerIntegrationTest` and `ParserIntegrationTest`

theofidry:
Many thanks!

PR #933, opened 2018-06-06
fix(denormalization): allow usage of dots in the ReferenceRangeNameDenormalizer

### LINKED ISSUE #932: Cannot use references with a dot in with the Ranged Denormalizer
Given yaml like

```
foo.bar.baz_{1..5}:
  foo: 'bar'

foo.bar.foz_{@foo.bar.baz_*}:
  bar: '<current()>'
```

This will return an error from the DenormalizerExceptionFactory:
```
No suitable fixture denormalizer found to handle the fixture with the reference "foo.bar.foz_{@foo.bar.baz_*}"
```

This is because the regex pattern [here](https://github.com/nelmio/alice/blob/50feea903db5c6e7be3ca465a9874d56a9a2f800/src/FixtureBuilder/Denormalizer/Fixture/Chainable/ReferenceRangeNameDenormalizer.php#L36) doesn't support dots.
 
Fix in #933

tomtomau:
This resolves #932

### DIFF
--- a/src/FixtureBuilder/Denormalizer/Fixture/Chainable/ReferenceRangeNameDenormalizer.php
+++ b/src/FixtureBuilder/Denormalizer/Fixture/Chainable/ReferenceRangeNameDenormalizer.php
@@ -33,7 +33,7 @@ final class ReferenceRangeNameDenormalizer implements ChainableFixtureDenormaliz
 {
     use IsAServiceTrait;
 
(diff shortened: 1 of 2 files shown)


theofidry:
Thanks for the fix! Could you add a small test for it as well? I believe a similar scenario to the one you described in the original issue in `LoaderIntegrationTest` would be enough

tomtomau:
Hey @theofidry - sorry for taking so long but I added a test case :)

theofidry:
Perfect, thanks! :)

Issue #936, opened 2018-06-20
unique constraint on self-referencing m:n entities causes stack trace


BigBadBassMan:
Hi there,

I'm using alice 3.4 with symfony 4 and hautelook/alice-bundle 2.0, so this seems to be recent:

I have several m:n relations like this (pseudo code, the goal is to have relations to several otherObj from one MyObj, but no duplicate entries in the relation table):

```
class MyObj {
    /**
     *
     * @var Collection
     *
     * @ORM\ManyToMany(targetEntity="OtherObj", inversedBy="relatedMys")
     * @ORM\JoinTable(name="my_other_obj",
     *      joinColumns={
     *          @ORM\JoinColumn(name="myObj", referencedColumnName="id")
     *      },
     *      inverseJoinColumns={
     *          @ORM\JoinColumn(name="otherObj", referencedColumnName="id", unique="true")
     *      }
     * )
     */
    private $relatedOthers;
}
```
'
when defining fixtures like this (with around 50 otherObj fixtures accordingly):
```
Namespace\MyObj:
    myobs{1..300}:
        relatedOthers (unique):  '<numberBetween(1, 10)>x @otherObj*
````

alice fails with a stack trace and the message: "Nesting level too deep - recursive dependency?"

dropping the unique constraint is no option, it is enforced by the database, and alice often seems to randomly choose duplicate values for a single related array, although there are plenty to choose from.

So the question is: am I doing something wrong, or is there some other way of achieving this? Lowering the numbers does not change a thing. Also defining static arrays of say 2 random related otherObj gives the same error.

theofidry:
It’s also possible. The best way would be to debug it, there is a
UniqueValueResolver or Pool which you can check

theofidry:
Ah I see, thanks for digging into this.

I'm a bit conflicted because I'm afraid there's cases where you expect comparison by _values_, i.e. you provide a pool of objects and you want to compare their values not their references but it will bail out on incomparable objects like you have

BigBadBassMan:
Thats why you're chief of staff ;-)

I cannot judge, if there are such cases, and if they are many.
Does your test suite cover this with the current comparison operator? It should bail on changing the operator, if some fundamentals break, somewhere.
(Maybe I can assist in extending the tests to includy my specific problem...)

I would really like to see this fix included.

theofidry:
>Thats why you're chief of staff ;-)

[...]

PR #938, opened 2018-07-03
uniquevaluespool: compare by reference 

BigBadBassMan:
As described in Issue https://github.com/nelmio/alice/issues/936, change the comparision operator to compare references only, not values to avoid infinite recursion on entities with cyclic/self-references.

### DIFF
--- a/src/Generator/Resolver/UniqueValuesPool.php
+++ b/src/Generator/Resolver/UniqueValuesPool.php
@@ -51,7 +51,7 @@ final class UniqueValuesPool
         }
 
         if (is_object($val1)) {
-            return $val1 == $val2;
+            return $val1 === $val2;
         }
 
         if (is_scalar($val1) || null === $val1) {
--- a/src/Generator/Resolver/UniqueValuesPool.php
+++ b/src/Generator/Resolver/UniqueValuesPool.php
@@ -50,8 +50,15 @@ final class UniqueValuesPool
             return false;
         }
 
-        if (is_object($val1)) {
-            return $val1 === $val2;
+        if (is_object($val1) && is_object($val2)) {
+            return $this->objectsAreIdentical($val1, $val2);
+        }
+
+        if (
+            (is_object($val1) && !is_object($val2))
+            || (!is_object($val1) && is_object($val2))
+        ) {
+            return false;
         }
 
         if (is_scalar($val1) || null === $val1) {
@@ -74,7 +81,7 @@ final class UniqueValuesPool
 
         return true;
     }
-    
+
     private function arrayHasValue($value, array $array): bool
     {
         foreach ($array as $arrayValue) {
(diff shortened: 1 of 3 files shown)


BigBadBassMan:
ok, not as easy as I tought. needs custom compare-function doing strict compare on all properties, not the objects itself.

theofidry:
> needs custom compare-function doing strict compare on all properties, not the objects itself.

I'm not sure to follow: either we compare two objects by equivalence or by reference and in the later it doesn't matter if they are equivalent or not

BigBadBassMan:
The thing is: your tests try to compare to separate instances of a class wich happen to have exactly the same property values:

```
$a = new StdClass();
$a->name = 'name';
$b = new StdClass();
$b->name = 'name';
```

This used to return true, since the comparison was by value. Comparing by reference fails this test, which can be rewritten. But I think the change in behaviour uncovers another "bug". Since identitity is now determined by reference, alice seems to create duplicate related entities - at least in my setup. Which is why I think that a simple by-reference comparison is not the solution.

src/Generator/Resolver/UniqueValuesPool.php
```diff
@@ -95,4 +102,43 @@ public function add(UniqueValue $value)
 
         $this->pool[$valueId][] = $value->getValue();
     }
+
+    private function objectsAreIdentical($o1, $o2): bool
```
theofidry: I would move this in a util class static method that could be unit tested separately

BigBadBassMan: Agreed, was just a quick fix to test something out. Care to suggest a name / location?

Also,  (and besides a some unit tests) this would need testing in more real-world applications, to get a view on the performance impact. And to make sure that no unintended behaviour changes occured.

Anyway: this allows me to use `(unique)` on m:n relations which may be cyclic, while not running into an infinite loop.

Any thoughts on the basic logic?

theofidry:
@BigBadBassMan what's the status of this PR?

BigBadBassMan:
@theofidry sorry for the delay, was on vacation and then got jumped by piled up work...
anyway, latest code passes all tests and works for me in my specific use-case.
I ended up offloading the comparison logic to [sebastian/comparator](https://github.com/sebastianbergmann/comparator), this is better tested and designed than anything I could come up with.
Unfortunately, this adds new dependencies - so if you do not want this, we could "reimplement" the comparison logic along those lines. 
Otherwise, some real-world testing would be in order, before merging this - from my point of view it is ready.

theofidry:
Thanks for the update :)

I'm fine with the changes and the new dependency, I would however require a test for it otherwise the change would both be undocumented and subject to regressions

BigBadBassMan:
> I'm fine with the changes and the new dependency, I would however require a test for it otherwise the change would both be undocumented and subject to regressions

Where/how would i implement this? currently, it is required in composer.json, thus beeing automatically resolved by composer. Do you have actual unit tests for libs alice is depending on?

theofidry:
Sorry I was confusing. I meant test for the change themselves not the added dependency per se. In #936 you showed there was a case of infinite recursion resulting in a fatal error before. I think you could add a test in `LoaderIntegrationTest` like [here](https://github.com/nelmio/alice/blob/master/tests/Loader/LoaderIntegrationTest.php#L3959). A good way to check if your test is correct is to put back the original code (the `==` comparison) and making sure it bails out and once you keep your code, the test should now be green

theofidry:
ping me on Slack if it's not clear enough (the link is in the readme)

BigBadBassMan:
I ended up adding an additional test method, not a simple generator entry as suggested.
This will fail with a PHP Fatal Error (which I found no way of capturing in PHPUnit), if the old comparison logic is still in place, otherwise it will pass.
Also, some exceptions needed to be catched, since the test may run into `alice`'s own depth limit of 150, and fail to generate objects - this is not related to my changes - it stems from the way `alice` picks samples from sets.

Aside from this, I consider this ready to be merged.

PR #943, opened 2018-08-11
Add more context on failure

### LINKED ISSUE #871: Make debugging easier
When an error occurs, it is quite hard to understand where it comes from.

I get an error like that:

```
  [Nelmio\Alice\Throwable\Exception\Generator\Resolver\FixtureNotFoundException]  
  Could not find the fixture "challenge_1A".    
```

But I have not idea where this `challenge_1A` reference comes from in my fixtures (AFAICT, I have no such reference).

### LINKED ISSUE #909: Fixture list syntax is not BC with Alice 2
In Alice 2, the separator for fixture lists was a comma followed by an optional space. In 3.x, the space is now required. This makes migration painful when fixtures were written without the space (the error message is totally understandable in this case btw).

Since the time I tried it, this has been documented in the upgrade guide, but it would still be great if the error message could be improved then.

theofidry:
Closes #871 
Closes #909

/cc @stof

### DIFF
--- a/src/Bridge/Symfony/Resources/config/fixture_builder/denormalizer/fixture.xml
+++ b/src/Bridge/Symfony/Resources/config/fixture_builder/denormalizer/fixture.xml
@@ -23,7 +23,12 @@
         </service>
 
         <service id="nelmio_alice.fixture_builder.denormalizer.fixture"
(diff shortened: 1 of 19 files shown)


stof:
CI is red, but this looks good to me.

PR #945, opened 2018-08-11
Add doc regarding extending non-template fixtures & fix the include order

### LINKED ISSUE #910: Extending a fixture coming from a fixture list does not work anymore in 3.x
In 2.x, the following would work, and merge settings of the fixtures (except the constructor arguments, which have to be repeated when extending):

```yaml

    challenge_{1A, 2A, 3A}:
        __construct: ['@galaxy_corp', '<{space_type.challenge}>']
        name: 'Prix <current()>'
        identifier: 'prix-<current()>'
        visibility: '<{space.visibility_public}>'
        administrators: ['@admin_user']
        individualTargetLevels:
            - '@agency_hierarchical_level'
        activate: []

    challenge_1A:
        __construct: ['@galaxy_corp', '<{space_type.challenge}>']
        name: 'Prix 1A / Agencies - All / Themed'
        identifier: 'prix-1a-team-only-themed'
        onlyTeamResults: true
        individualTargetLevels: []
        hierarchicalLevels:
            - '@agency_hierarchical_level'
        theme: 'sell_sell_sell'
```

In 3.x, this does not work anymore. Only settings from the explicit fixture are taken into account (I think it takes the last one into account so the one from the fixture list could win by swapping them). But this is not documented in the upgrade guide.

### LINKED ISSUE #941: NativeLoader->LoadFiles load data in reverse order
We add numbers at start of the filenames to ensure that the loading order is as expected.

Upgrading from 1.5.*@dev to v3.1.3 introduce some bigs changes.

Files data are added in database in reverse order...

To fix it, we have modified this script on line 38: `vendor\nelmio\alice\src\Parser\IncludeProcessor\IncludeDataMerger.php`

Before: 
```php
? array_merge($includeData[$class], $fixtures)            
````

After:
```php
 ? array_merge($fixtures, $includeData[$class])
````

### PR DESCRIPTION
**Pull request by theofidry**
Closes #910
Closes #941

/cc @sami-bel @stof

### DIFF
--- a/src/Parser/IncludeProcessor/DefaultIncludeProcessor.php
+++ b/src/Parser/IncludeProcessor/DefaultIncludeProcessor.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Nelmio\Alice\Parser\IncludeProcessor;
 
+use function array_reverse;
 use Nelmio\Alice\FileLocatorInterface;
 use Nelmio\Alice\IsAServiceTrait;
 use Nelmio\Alice\Parser\IncludeProcessorInterface;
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/Parser/IncludeProcessor/DefaultIncludeProcessor.php [hunk 1]
```diff
@@ -75,6 +76,42 @@ public function process(ParserInterface $parser, string $file, array $data): arr
 
         $this->included[$file] = true;
 
+        // The order the the include statements needs to be reversed. Indeed when merging two sets of data, e.g.
```
theofidry: That change of order needs to be documented. The current behaviour in 3.x is in line with 2.x but I think it's a bug and doesn't make sense from a user PoV. Bug that would be a BC break still (which is worth fixing IMO since it's a bug)

PR #946, opened 2018-08-11
Parse tokens as regular strings when unrecognised

### LINKED ISSUE #913: Escaping of % is incompatible between 2.x and 3.x
When wanting to put a literal `%` in a string, following a number, the syntax is different between 2.x and 3.x:

- in 2.x, it CANNOT be escaped (putting a backslash before `%` would also put it in the final value)
- in 3.x, it MUST be escaped, to prevent a syntax error (with an unclear message)

```diff
-        name: '0C - 1% du Chiffre d''affaire'
+        name: '0C - 1\% du Chiffre d''affaire'
```

theofidry:
Closes #913


stof:
@theofidry any idea when this will be released ?

PR #950, opened 2018-09-01
Fix comparison of objects

### LINKED ISSUE #936: unique constraint on self-referencing m:n entities causes stack trace

theofidry:
Closes #938
Closes #936

Rebased & reviewed version of #938.

When comparing objects for unique values, a fatal error was thrown when doing `$object1 == $object2` whenever there was a circular reference.

Changes compared to #938:

- Retouched the unique ID generated for `UniqueValueDenormalizer` (unrelated change but saw that at the same time)
- Use a single instance of `SebastianBergmann\Comparator\Factory` instead of instantiating it each time
- Rewrote the test to ensure a fatal error is occurring before this patch

/cc @BigBadBassMan

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "php": "^7.1",
         "fzaninotto/faker": "^1.6",
         "myclabs/deep-copy": "^1.5.2",
+        "sebastian/comparator": "^3.0",	
         "symfony/property-access": "^2.8 || ^3.4 || ^4.0",
         "symfony/yaml": "^2.8 || ^3.4 || ^4.0"
     },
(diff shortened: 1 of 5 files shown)


theofidry:
Thank you very much @BigBadBassMan

PR #953, opened 2018-09-10
Remove obsolete article from the Readme

Einenlum:
This article is nice but it is about Alice 2.x and not 3.x.
It should be removed from the Readme.


theofidry:
Thanks for checking it

PR #955, opened 2018-09-20
Fix: EvaluatedValueResolver fails to resolve current with range references

goetas:
EvaluatedValueResolver fails to resolve current with range references.

The feature works as expected with `<current()>` but not with `<($current)>`

### DIFF
--- a/src/Generator/Resolver/Value/Chainable/EvaluatedValueResolver.php
+++ b/src/Generator/Resolver/Value/Chainable/EvaluatedValueResolver.php
@@ -58,6 +58,10 @@ final class EvaluatedValueResolver implements ChainableValueResolverInterface
         $_scope = $scope;
         try {
             $_scope['current'] = $fixture->getValueForCurrent();
+
+            if ($_scope['current'] instanceof FixtureInterface) {
+                $_scope['current'] = $fixtureSet->getObjects()->get($_scope['current'])->getInstance();
+            }
         } catch (NoValueForCurrentException $exception) {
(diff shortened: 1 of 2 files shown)


src/Generator/Resolver/Value/Chainable/EvaluatedValueResolver.php
```diff
@@ -58,6 +58,10 @@ public function resolve(
         $_scope = $scope;
         try {
             $_scope['current'] = $fixture->getValueForCurrent();
+
+            if ($_scope['current'] instanceof FixtureInterface) {
+                $_scope['current'] = $fixtureSet->getObjects()->get($_scope['current'])->getInstance();
```
theofidry: what about setting the right value directly L60?

goetas: With using a ternary operator? Do not understand

theofidry:
Yep with a tenary operator

goetas:
Does not sound a good idea as I will have to call `$fixture->getValueForCurrent();` twice...

```php
 $_scope['current'] = $fixture->getValueForCurrent() FixtureInterface 
    ? $fixture->getValueForCurrent()
    : $fixtureSet->getObjects()->get($fixture->getValueForCurrent())->getInstance();
```

theofidry:
Fair enough, thanks for the patch!

goetas:
Great! Thanks!

PR #963, opened 2018-11-08
[Failing-test] The order of how files are loaded should not be important

goetas:
The order of how files are loaded should not be important, but it look that when having references to other objects, things are not so..

### DIFF
--- a/src/FixtureBuilder/Denormalizer/Fixture/SimpleFixtureBagDenormalizer.php
+++ b/src/FixtureBuilder/Denormalizer/Fixture/SimpleFixtureBagDenormalizer.php
@@ -58,7 +58,7 @@ final class SimpleFixtureBagDenormalizer implements FixtureBagDenormalizerInterf
     {
         $fixtures = new FixtureBag();
 
-        $alreadyRetired = [];
+        $alreadyRetried = [];
         while ($result = array_splice($data, 0, 1)) {
             $fqcnWithFlags = key($result);
             $rawFixtureSet = current($result);
@@ -89,12 +89,12 @@ final class SimpleFixtureBagDenormalizer implements FixtureBagDenormalizerInterf
                         $flags
                     );
                 }
-            } catch (UnexpectedValueException $e) {
-                if (!isset($alreadyRetired[$fqcnWithFlags])) {
-                    $data [$fqcnWithFlags] = $rawFixtureSet;
-                    $alreadyRetired[$fqcnWithFlags] = true;
+            } catch (UnexpectedValueException $exception) {
+                if (!isset($alreadyRetried[$fqcnWithFlags])) {
+                    $data[$fqcnWithFlags] = $rawFixtureSet;
+                    $alreadyRetried[$fqcnWithFlags] = true;
                 } else {
-                    throw $e;
+                    throw $exception;
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


goetas:
The solution used to sole this is to delay the creation of some fixtures in case of errors

src/FixtureBuilder/Denormalizer/Fixture/SimpleFixtureBagDenormalizer.php
```diff
@@ -56,7 +57,12 @@ public function __construct(FixtureDenormalizerInterface $fixtureDenormalizer, F
     public function denormalize(array $data): FixtureBag
     {
         $fixtures = new FixtureBag();
-        foreach ($data as $fqcnWithFlags => $rawFixtureSet) {
+
+        $alreadyRetired = [];
```
theofidry: `retried` instead of `retired`?

src/FixtureBuilder/Denormalizer/Fixture/SimpleFixtureBagDenormalizer.php
```diff
@@ -69,22 +75,30 @@ public function denormalize(array $data): FixtureBag
...
-                    $specs = [];
+                    $fixtures = $this->fixtureDenormalizer->denormalize(
+                        $fixtures,
+                        $fqcn,
+                        $reference,
+                        $specs ?? [],
+                        $flags
+                    );
+                }
+            } catch (UnexpectedValueException $e) {
```
theofidry: just for consistency sake with the rest of the codebase I would keep `$exception` instead of `$e` here

src/FixtureBuilder/Denormalizer/Fixture/SimpleFixtureBagDenormalizer.php
```diff
@@ -69,22 +75,30 @@ public function denormalize(array $data): FixtureBag
...
+                if (!isset($alreadyRetired[$fqcnWithFlags])) {
+                    $data [$fqcnWithFlags] = $rawFixtureSet;
```
theofidry: is the space between `$data` and `[$fqcnWithFlags]` intentional here?

src/FixtureBuilder/Denormalizer/Fixture/SimpleFixtureBagDenormalizer.php
theofidry: From where is this exception thrown?

goetas: The exception is triggered inside the  `get` call in https://github.com/nelmio/alice/blob/19bec34576564cb54921e096bf0e5eba2bee027e/src/FixtureBuilder/Denormalizer/Fixture/Chainable/ReferenceRangeNameDenormalizer.php#L130 and is the core of this PR. 

When `get` is called, the necessary fixture to reference has not yet been loaded, with the `while` loop introduced in this PR, we retry once more at the end of the loading process.

goetas:
@theofidry all issues from your last comments have been addressed.

theofidry:
Thanks a lot @goetas!

Just a thought though: what happens if you put the wrong order within the same file?

goetas:
Hmm, that's a good question. Will check it and let you know.

goetas:
> Just a thought though: what happens if you put the wrong order within the same file?

See https://github.com/nelmio/alice/pull/994

Issue #569, opened 2016-09-23
v3 - HTML code in fixtures


mcjwb:
Hi,
Is there a way to allow HTML in fixtures (or disable the lexer for certain fixtures)?
On my first attempt I got the error:
`PHP Fatal error:  Uncaught Nelmio\Alice\Exception\FixtureBuilder\ExpressionLanguage\LexException: Could not lex the value...`
Then on a simplified test case I got the error:

```
PHP Fatal error:  Uncaught InvalidArgumentException: Invalid token "&lt;div&gt;test&lt;/div&gt;" found. in /var/www/project/vendor/nelmio/alice/src/FixtureBuilder/ExpressionLanguage/Lexer/SubPatternsLexer.php:100
Stack trace:
# 0 /var/www/project/vendor/nelmio/alice/src/FixtureBuilder/ExpressionLanguage/Lexer/SubPatternsLexer.php(76): Nelmio\Alice\FixtureBuilder\ExpressionLanguage\Lexer\SubPatternsLexer->lexFragment(Object(Nelmio\Alice\FixtureBuilder\ExpressionLanguage\Lexer\ReferenceLexer), '&lt;div&gt;test&lt;/div&gt;')
# 1 /var/www/project/vendor/nelmio/alice/src/FixtureBuilder/ExpressionLanguage/Lexer/FunctionLexer.php(52): Nelmio\Alice\FixtureBuilder\ExpressionLanguage\Lexer\SubPatternsLexer->lex('&lt;div&gt;test&lt;/div&gt;')
# 2 /var/www/project/vendor/nelmio/alice/src/FixtureBuilder/ExpressionLanguage/Lexer/LexerRegistry.php(45): Nelmio\Alice\FixtureBuilder\ExpressionLanguage\Lexer\FunctionLexer->lex('&lt;div&gt;test&lt;/div&gt;')
# 3 /var/www/project/vendor/nelmio/alice/src/FixtureBuilder/ExpressionLanguage/Parser/SimpleParser.php(50): Nelmio\Alice\FixtureBuilder\ExpressionLanguage\Lexer\Lexer in /var/www/project/vendor/nelmio/alice/src/FixtureBuilder/ExpressionLanguage/Lexer/SubPatternsLexer.php on line 100
> 
[...]

PR #967, opened 2018-12-10
Allow line breaks in function parameters

tigitz:
Hello,

As recommended here: https://github.com/nelmio/alice/issues/569#issuecomment-299711055 I wanted to implement something like:

```yaml
Fixtures\Data:
    emails:
        body: "<identity('<p>foo\n\rbar</p>')"
```

But the function regex parser used the `.` token to extract parameters which excludes the line breaks.

I added `\n` and `\r` in addition to the `.` token to make it work with those kind of line breaks. I'm still new to alice so I might have missed some usage that could be broken with these additions. Let me know if this is the case.

I considered the `\X` token but It seems to widen the scope more than it's needed for my use case. Let me know if it it makes more sense to you.

Thanks

### DIFF
--- a/tests/FixtureBuilder/ExpressionLanguage/Lexer/SubPatternsLexerTest.php
+++ b/tests/FixtureBuilder/ExpressionLanguage/Lexer/SubPatternsLexerTest.php
@@ -48,6 +48,19 @@ class SubPatternsLexerTest extends TestCase
         $this->assertEquals($expected, $actual);
     }
 
+    public function testLexAFunctionContainingLineBreaks()
+    {
+        $expected = [
+            new Token('<identity("foo'.PHP_EOL.'bar")>', new TokenType(TokenType::FUNCTION_TYPE)),
+        ];
+
+        $lexer = new SubPatternsLexer(new FakeLexer());
+        $actual = $lexer->lex('<identity("foo'.PHP_EOL.'bar")>');
+
+        $this->assertCount(count($expected), $actual);
+        $this->assertEquals($expected, $actual);
+    }
+
     public function testUsesDecoratedLexerToLexReferenceValues()
     {
         $value = '@user';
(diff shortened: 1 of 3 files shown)


tests/FixtureBuilder/ExpressionLanguage/Lexer/SubPatternsLexerTest.php
```diff
@@ -48,6 +48,19 @@ public function testLexAValueToReturnAListOfTokens()
         $this->assertEquals($expected, $actual);
     }
 
+    public function testLexAFunctionContainingLineBreaks()
+    {
+        $expected = [
+            new Token('<identity("foo'.PHP_EOL.'bar")>', new TokenType(TokenType::FUNCTION_TYPE)),
```
theofidry: I think it would be better to use `\n` in one case and `\r\n` for Windows

tigitz: Alright, to be clear does that mean I should do 2 separate test functions for both `\n` and `\r\n` case ? Should I check the OS to make sure it's Windows before running the `\r\n` case ? Also it seems that mac has `\r` line breaks should I update the regex to this : https://stackoverflow.com/a/1331840 ?

tigitz:
Sure, I will also implement those tests as soon as the proper regex is defined :+1:

tests/FixtureBuilder/ExpressionLanguage/Lexer/SubPatternsLexerTest.php
theofidry: I think you got an answer regarding the regex on StackOverflow.

You can either go with 2 separate test functions or only one with a data provider, I'm fine with either

tigitz:
@theofidry tests added and regex improved.

I thought about implementing tests for double quoted string containing something like `<function("foo\nbar")>` to make sure the argument value is equal to:
```
foo

bar
```

but that would be OS dependent, on windows it would output `foo\nbar`.

I could put some OS checks before yielding the expected value if we want to go for this kind of test.

wdyt ?

theofidry:
If that's easily doable then yes. If not I think the current patch is good as it is: if windows users feel differently then they would need to step up and help out with the windows support :)

tigitz:
Hey @theofidry, when you have some time it would be cool if you could check the PR. I think it's in the state where it could be merged.

Thank you :wink:

PR #970, opened 2019-01-12
fix document on expression language

tzmfreedom:
Fix docs on expression language. It should show bound parameter.


theofidry:
Thanks!

PR #976, opened 2019-02-28
Changed use of self to static to allow for easier extension

Bittarman:
Changed FakerGeneratorFactory::create calls use of LOCALE constant from self to static to allow for easier overriding


theofidry:
Thanks

PR #979, opened 2019-03-12
Reference range flags (e.g. template) fix

daniel-iwaniec:
When using reference ranges templating doesn't work (possibly other flags too).
This change is analogous to `SimpleDenormalizer`.


daniel-iwaniec:
Currently I'm forced to implement this class by myself in my project and override `\Nelmio\Alice\Loader\NativeLoader::createFixtureDenormalizer`.

theofidry:
Thanks for the PR! Do you mind adding a test for it as well to make sure there won't be a regression in the future?

daniel-iwaniec:
There wasn't any tests for `ReferenceRangeNameDenormalizer` at all so I didn't mind not covering it.
I've added `ReferenceRangeNameDenormalizerTest` similarly to `SimpleDenormalizerTest` with some basic tests as well as specific test for adding template flag to fixture.

theofidry:
Thanks a lot! Sorry I was not clear enough: whilst the unit test is really nice (and really good job for it!), I would really like to see an integration test for it. Indeed it's a component for which how it's configured matters a lot.

The best way to go I think would be adding a in [this test](https://github.com/nelmio/alice/blob/master/tests/Loader/LoaderIntegrationTest.php#L3989-L3990).

daniel-iwaniec:
I've added integration test you've asked for.

theofidry:
Thanks very muchç

PR #980, opened 2019-03-13
Fix lexer on long arguments

jderusse:
#967 introduced a regression when argument are too longs. 

When the string is very long, the regexp engine is fine with `.*` but isn't with `(.|.)*`


theofidry:
Thank you very much :)

PR #983, opened 2019-03-27
Update README.md

tolstoyfafa:


theofidry:
Isn't the current wording more correct?

snebes:
the phrasing "head here" and "read hear" are both a but unclear.  What do you think of "browse the alice 2.x documentation"?

theofidry:
Maybe it can be reworded as "If you want to check the documentation for 2.x, click [this link]"?

Issue #973, opened 2019-02-19
randomElement and named relations


pbowyer:
Alice has support for picking a random item using the wildcard syntax: `@element*`. This works great when I've generated fixtures using a loop, ending up with `@element1`, `@element2` ... `@elementN`.

Where values aren't random (e.g. settings, statuses) and I manually create the fixtures, I give them descriptive names so I can easily reference them in other manually created fixtures. E.g.:
```
App\Entity\Status:
    success:
        ...
    failure: # Easier to remember than @status2
        ...
```

But in the generated fixtures, I need to pick a random status.

I have tried:
`<randomElement([@success, @failure])>`, `<randomElement(['@success', '@failure'])>`, defining a parameter that's an array of the statuses; all give the error:
```
Expected argument of type "App\Entity\Setting\Status or null", "Nelmio\Alice\Definition\Value\FixtureReferenceValue" given at property path "status".
```
Is it possible to pick a random element from a user-defined set?

theofidry:
It should be possible, that's a proper bug that needs to be solved

pifaace:
I confirm, I created a custom provider and a fixture like this : 
```yaml
validationGroup1:
      validationGroupId: 'validationGroup1'
      name: 'sweet validatonGroup1'
      persons: <personCollection([@johnDoe, @janeDoe])>
      orderables: '<orderableCollection([@flyer])>'
```
and the arg in my personCollection method is an array of "FixtureReferenceValue". It doesn't convert the references to object

theofidry:
Closed by #984

PR #984, opened 2019-04-01
fix(): Fix the missing resolving or references in simple string arrays

TiMESPLiNTER:
I had the problem that references in an array:

```
['@user1', '@user2']
```

were correctly recognized but never resolved to a `User` object but stayed as a reference object in the data structure. This PR fixes this.

### DIFF
--- a/src/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/FunctionTokenParser.php
+++ b/src/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/FunctionTokenParser.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 
 namespace Nelmio\Alice\FixtureBuilder\ExpressionLanguage\Parser\TokenParser\Chainable;
 
-use Nelmio\Alice\Definition\Value\ArrayValue;
 use Nelmio\Alice\Definition\Value\EvaluatedValue;
 use Nelmio\Alice\Definition\Value\FunctionCallValue;
 use Nelmio\Alice\Definition\Value\ValueForCurrentValue;
@@ -101,7 +100,6 @@ final class FunctionTokenParser implements ChainableTokenParserInterface, Parser
         } elseif ('current' === $function) {
             $arguments = [new ValueForCurrentValue()];
         } else {
-            /** @var ArrayValue $arguments */
             $arguments = $this->parseArguments($this->parser, trim($matches['arguments']));
         }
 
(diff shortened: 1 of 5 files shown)


theofidry:
Thanks very much for the fix! Could you add a test as well to [LoaderIntegrationTest](https://github.com/nelmio/alice/blob/master/tests/Loader/LoaderIntegrationTest.php#L4027)? Your patch should be enough strictly speaking but I prefer to add an extra safety with an integration test

TiMESPLiNTER:
@theofidry isn't this the test already? https://github.com/nelmio/alice/blob/master/tests/Loader/LoaderIntegrationTest.php#L2546

It looks like exactly what I needed. Just asking myself why this test passes in the current master. 🤔

theofidry:
Ah indeed, then could it be that the reproducer is not just having a plain array? Do you have a unique property or something with it?

TiMESPLiNTER:
```
 'another_dummy' => [
    'dummies' => ['@dummy', '@dummy', '@dummy'],
],
```

this is a plain array isn't it? Or maybe I don't get exactly what you mean by "plain array"...

theofidry:
>this is a plain array isn't it? Or maybe I don't get exactly what you mean by "plain array"...

Yes it is.

There is several reports (like https://github.com/nelmio/alice/issues/973) which hints there is an issue with some values which are not properly resolved, so I don't doubt you encountered that situation. I just think the failing scenario is not the one you original described. But maybe the other linked issue can give an idea of what can cause this failure

TiMESPLiNTER:
@theofidry yes, issue #973 was exactly the problem I solved with this PR now. So #973 will also be fixed once this gets merged. It must be something different as the test I mentioned passes in master without problems. Although I can't come up with another idea which constellation reproduces the issues mentioned here and in #973. So I know what causes the failure described there and this PR fixes it.

theofidry:
Could you try:

```php
yield 'array value' => [
            [
                stdClass::class => [
                    'dummy' => [
                        'foo' => 'bar',
                    ],
                    'another_dummy' => [
                        'dummies' => '[@dummy, @dummy, @dummy]',
                    ],
                ],
            ],
            [
                'parameters' => [],
                'objects' => [
                    'dummy' => $dummy = StdClassFactory::create([
                        'foo' => 'bar',
                    ]),
                    'another_dummy' => StdClassFactory::create([
                        'dummies' => [$dummy, $dummy, $dummy]
                    ]),
                ],
            ],
        ];
```

or:

```php
yield 'array value' => [
            [
                stdClass::class => [
                    'dummy' => [
                        'foo' => 'bar',
                    ],
                    'another_dummy' => [
                        'dummies' => '<randomElement([@dummy, @dummy, @dummy])>',
                    ],
                ],
            ],
            [
                'parameters' => [],
                'objects' => [
                    'dummy' => $dummy = StdClassFactory::create([
                        'foo' => 'bar',
                    ]),
                    'another_dummy' => StdClassFactory::create([
                        'dummies' => [$dummy, $dummy, $dummy]
                    ]),
                ],
            ],
        ];
```

without your patch first to see it if makes it fail and then with your patch to see if it solves the issue?

TiMESPLiNTER:
Nice the first one works. Fails without my change and passes with my change. 👍 I just pushed it into this branch.

theofidry:
Thanks!

PR #994, opened 2019-05-05
Add extra test when the fixtures are in the same file but different order

goetas:
This PR addressees https://github.com/nelmio/alice/pull/963#issuecomment-489399545

> Just a thought though: what happens if you put the wrong order within the same file?

The problem was the same even if the fixtures were in the same file but in different order, but This PR addressees https://github.com/nelmio/alice/pull/963 fixes the issue.

I have added an extra test to make sure to not have regressions.


goetas:
The idea was to make it explicit in the test that the order is not
important, thus the two files... Even if you the same case is present in
other tests

theofidry:
Yeah true, I'm confusing with another case which that's not part of this problem but about the fixtures reference feature itself instead.

theofidry:
Thanks!

PR #1000, opened 2019-08-02
Composer: exclude some folders from downloading into vendor

f3l1x:
Hi. This should exlude these folders from composer install.

image


f3l1x:
Wow, hit 1000 issue number, cool. :-)

theofidry:
Thanks

PR #1008, opened 2019-11-21
Added Symfony 5 support

koenreiniers:
Symfony 4.4 and 5.0 were released today, so I decided to add support for both of them.

I had to enable the serializer since the new error handler component depends on it.

I've never worked with travis CI before, so please forgive me if I did something incorrect there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,8 @@ matrix:
       env: SYMFONY_VERSION='~4.0.0'
     - php: '7.2'
       env: SYMFONY_VERSION='~4.1.0@dev'
+    - php: '7.2'
+      env: SYMFONY_VERSION='~5.0.0'
   allow_failures:
     - php: nightly
     - env: SYMFONY_VERSION='~4.1.0@dev'
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,8 @@ matrix:
     - php: '7.2'
       env: SYMFONY_VERSION='~4.1.0@dev'
     - php: '7.2'
+      env: SYMFONY_VERSION='~4.4.0'
+    - php: '7.2'
       env: SYMFONY_VERSION='~5.0.0'
   allow_failures:
     - php: nightly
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -26,6 +26,10 @@ matrix:
       env: SYMFONY_VERSION='~4.0.0'
     - php: '7.2'
       env: SYMFONY_VERSION='~4.1.0@dev'
```
theofidry: I think we can remove both the 4.1.0@dev and 4.0.0 one. The 4.4.0 one should be enough

.travis.yml
```diff
@@ -26,6 +26,10 @@ matrix:
...
+    - php: '7.2'
+      env: SYMFONY_VERSION='~4.4.0'
+    - php: '7.2'
+      env: SYMFONY_VERSION='~5.0.0'
   allow_failures:
     - php: nightly
     - env: SYMFONY_VERSION='~4.1.0@dev'
```
theofidry: as a result of my previous comment: this one should be removed as well

composer.json
```diff
@@ -23,17 +23,17 @@
         "php": "^7.1",
         "fzaninotto/faker": "^1.6",
         "myclabs/deep-copy": "^1.5.2",
-        "sebastian/comparator": "^3.0",	
-        "symfony/property-access": "^2.8 || ^3.4 || ^4.0",
-        "symfony/yaml": "^2.8 || ^3.4 || ^4.0"
+        "sebastian/comparator": "^3.0",
+        "symfony/property-access": "^2.8 || ^3.4 || ^4.0 || ^5.0",
```
theofidry: I'm starting to irk at the idea we are going as far back as 2.8; maybe in another PR it could be wise to drop PHP 7.1 for minimum PHP 7.2 and drop Symfony 2.x and 3.x as well

koenreiniers: I think 2.8 is being EOL'd at the end of the month, so that is probably a good idea yeah. I think 3.4 should still be supported though since it's LTS until end of 2021 and a lot of people will probably still be using it. (I know at my company we are).

theofidry: If it doesn't cost anything for supporting the 3.4 then cool otherwise it can be dropped: the old version will remain available

theofidry:
Thank you @koenreiniers!

PR #1021, opened 2020-03-05
Update the list of PHP versions on CI

stof:
PHP 7.3 and 7.4 are added. nightly is removed as this is now 8.0-dev and so the job does not even start.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ php:
 
 env:
   global:
-    - PHPUNIT_FLAGS='--stop-on-failure --verbose'
+    - PHPUNIT_FLAGS='--verbose'
 
 matrix:
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,6 @@ matrix:
       env: SYMFONY_VERSION='~5.0.0'
 
 before_install:
-  - set -eo pipefail
   - echo "memory_limit=-1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
   - phpenv config-rm xdebug.ini || true
   - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,8 @@ cache:
 php:
   - '7.1'
   - '7.2'
-  - nightly
+  - '7.3'
+  - '7.4'
 
 env:
   global:
@@ -26,8 +27,6 @@ matrix:
       env: SYMFONY_VERSION='~4.4.0'
     - php: '7.2'
       env: SYMFONY_VERSION='~5.0.0'
-  allow_failures:
-    - php: nightly
 
 before_install:
   - set -eo pipefail
--- a/tests/Definition/Value/UniqueValueTest.php
+++ b/tests/Definition/Value/UniqueValueTest.php
@@ -89,7 +89,13 @@ class UniqueValueTest extends TestCase
         $this->assertEquals('(unique) \'foo\'', (string) $value);
 
         $value = new UniqueValue('', new \stdClass());
-        $this->assertEquals("(unique) stdClass::__set_state(array(\n))", (string) $value);
+
+        if (\PHP_VERSION_ID >= 70300) {
+            $expectedStdClass = "(unique) (object) array(\n)";
+        } else {
+            $expectedStdClass = "(unique) stdClass::__set_state(array(\n))";
+        }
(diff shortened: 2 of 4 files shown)


stof:
I'm also fixing existing tests to bring the CI green.

The error `Error: Call to undefined method phpDocumentor\Reflection\DocBlock\Tags\InvalidTag::getMethodName()` is a bug in prophecy, which is patched at https://github.com/phpspec/prophecy/pull/468

stof:
And the testsuite is green :tada:

.travis.yml
```diff
@@ -7,11 +7,12 @@ cache:
...
   - '7.1'
   - '7.2'
-  - nightly
+  - '7.3'
+  - '7.4'
 
 env:
   global:
-    - PHPUNIT_FLAGS='--stop-on-failure --verbose'
+    - PHPUNIT_FLAGS='--verbose'
```
theofidry: is there any reason to prevent a fail fast anymore?

.travis.yml
```diff
@@ -26,11 +27,8 @@ matrix:
       env: SYMFONY_VERSION='~4.4.0'
     - php: '7.2'
       env: SYMFONY_VERSION='~5.0.0'
-  allow_failures:
-    - php: nightly
 
 before_install:
-  - set -eo pipefail
```
theofidry: without it, it means you can have an error from `covers-validator` for example but still carrying on. I'm not sure that's a good idea

.travis.yml
stof: well, as I explained in the commit message, I find it more useful for a CI to report *all* failing tests than to stop the build fast. It gives you a chance to fix them all before resubmitting rather than having to do it one by one (and it allows maintainers to have an idea about how broken the tests are).

.travis.yml
stof: that's not true. `covers-validator` is in the `install` section, so a failure there already mark the build as being in the error state and stops it.
But having this here actually breaks Travis: a failing command in the `script` section was not properly marking the build as being in the `failed` state (and if you were trying to use `after_failure`, it would not work either).

On a side node, I think that `covers-validator` should be a `script` step rather than an `install` step (considering it one of the test commands triggering the `failed` state but letting other tests run). To me, it is perfectly fine to run your testsuite even if a `@covers` annotation is missing so that you get the full feedback from the CI.

.travis.yml
theofidry: I would prefer a fail fast tbh: if tests are broken you are likely going to need to run `make test` to ensure everything works fine and it relieves a bit the CI which is not always the fastest

.travis.yml
theofidry: >To me, it is perfectly fine to run your testsuite even if a @covers annotation is missing so that you get the full feedback from the CI.

It's fine when you are in the logic of letting the CI give you all the mistakes, but it's not really the logic I'm going for.

> But having this here actually breaks Travis: a failing command in the script section was not properly marking the build as being in the failed state (and if you were trying to use after_failure, it would not work either).

Interesting to know... So what is the proper way to fail fast on a multi-step scripts?

>I think that covers-validator should be a script step rather than an install step 

Agreed, not sure why it's there

.travis.yml
stof: well, I might have green tests locally, because only some PHP versions are broken.

and the complete PHPUnit run takes 1.19s. I don't think a faster failure will change much for the CI.
If you want CI to go faster, a better way would be to avoid the 1min spent in `composer bin all update` (while not all tools are needed by each job, and `covers-validator` could be run in a single job rather than all as well to avoid the need in others). Btw, the failed resolution for `humbug` is probably a culprit there.

.travis.yml
stof: well, the `scripts` section is precisely meant to run all the various scripts, so that you get all your test failures in the CI output.
If you want it to stop the build, keep it in `install`

stof: https://travis-ci.org/nelmio/alice/builds/658807997 proves it keeps breaking the build btw (and it now properly renders the Travis message because Travis ends the build cleanly rather than `pipefail` making the build script crash)

theofidry: Let's remove it then

.travis.yml
theofidry: Yeah I need to review a bit the bundle and bump some dependencies :)

theofidry:
Thank you @stof! :)

PR #1023, opened 2020-03-06
Make testsuite compatible with PHPUnit 8

stof:
PHPUnit 7 is unmaintained and does not officially support PHP 7.4.
Given that the testsuite still needs to run on PHPUnit 7 for the tests on PHP 7.1, I kept it compatible with both PHPunit 7 and PHPUnit 8.

Here is the summary of changes:

- added `void` return types on hook methods (this is compatible with PHPUnit 7, as child classes are allowed to add a return type when overriding methods)
- migrate from `@expectedException` to `expectException()` as the annotation-based expectations are deprecated in PHPUnit 8
- remove broken data provider est case due to tests typehinting the argument as string (PHPUnit 8 runs in strict mode, so the integer was not casted to string implicitly making it the same case than the previous one)
- fix invalid data providers (PHPUnit 8 reports them instead of silently ignoring duplicate keys like older versions which were using `iterator_to_array`), which allowed discovering some non-executed cases

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 /vendor-bin/*/composer.lock
 /humbuglog.json
 /humbuglog.txt
+/.phpunit.result.cache
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "bamarni/composer-bin-plugin": "^1.1.0",
         "php-mock/php-mock": "^2.0",
         "phpspec/prophecy": "^1.6",
-        "phpunit/phpunit": "^7.0",
+        "phpunit/phpunit": "^7.0 || ^8.5",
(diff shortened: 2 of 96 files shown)


theofidry:
@stof what about removing PHP 7.1 & 7.2? That would simplify things no?

stof:
@theofidry indeed, dropping 7.1 would simplify it (especially when it comes to supporting PHPUnit 9 as well, which was my initial goal, as supporting both 7 and 9 requires dealing with `expectExceptionMessageRegExp` being removed in 9). But I did not want to force such change of supported version yet (that's up to you for deciding).

theofidry:
Let's do it then. It's still on 7.1 by pure laziness on my part not by need :)

theofidry:
Merging this already since it works fine :) Thanks for the PR @stof!

PR #1024, opened 2020-03-06
Add support for sebastian/comparator 4

### LINKED ISSUE #1017: Requirement on sebastian/comparator:^3.1 blocks phpunit:^9.0
The root requirement of `sebastian/comparator:^3.0` blocks the usage of `phpunit/phpunit:^9.0` as it requires `sebastian/exporter:^4.0`.

The dependency tree of `sebastia/comparator:^3.0` requires `sebastian/exporter:^3.1`.
But using `sebastian/comparator:^4.0` means dropping support for `php:^7.1` and `php:^7.2`

With semver this will not require a new major version of alice itself.
But this update should be considered in the next release.

stof:
The CI always uses version 3 for now, because of the PHPUnit version being used (using sebastian/comparator 4 would require allowing the testsuite to run with PHPUnit 9, which is more work).
Given that the *only* change in sebastian/comparator 4 is the minimum PHP requirement (there are also some tooling changes in the git history, but that's only for contributors), I consider that this is fine as a way to fix #1017

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "php": "^7.1",
         "fzaninotto/faker": "^1.6",
         "myclabs/deep-copy": "^1.5.2",
-        "sebastian/comparator": "^3.0",
+        "sebastian/comparator": "^3.0 || ^4.0",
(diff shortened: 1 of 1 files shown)


theofidry:
Thanks

PR #1026, opened 2020-04-03
Add named parameters support

julienfalque:
Implements #594.

### DIFF
--- a/src/Generator/Instantiator/Chainable/NoCallerMethodCallInstantiator.php
+++ b/src/Generator/Instantiator/Chainable/NoCallerMethodCallInstantiator.php
@@ -15,9 +15,21 @@ namespace Nelmio\Alice\Generator\Instantiator\Chainable;
 
 use Nelmio\Alice\Definition\MethodCall\NoMethodCall;
 use Nelmio\Alice\FixtureInterface;
+use Nelmio\Alice\Generator\NamedArgumentsResolver;
 
 final class NoCallerMethodCallInstantiator extends AbstractChainableInstantiator
 {
+    /**
+     * @var NamedArgumentsResolver|null
+     */
+    private $namedArgumentsResolver;
+
+    // TODO: make $namedArgumentsResolver non-nullable in 4.0. It is currently nullable only for BC purposes
+    public function __construct(NamedArgumentsResolver $namedArgumentsResolver = null)
+    {
+        $this->namedArgumentsResolver = $namedArgumentsResolver;
+    }
+
     /**
      * @inheritDoc
      */
@@ -35,9 +47,13 @@ final class NoCallerMethodCallInstantiator extends AbstractChainableInstantiator
     {
         list($class, $arguments) = [
             $fixture->getClassName(),
-            array_values($fixture->getSpecs()->getConstructor()->getArguments())
+            $fixture->getSpecs()->getConstructor()->getArguments()
         ];
 
-        return new $class(...$arguments);
+        if (null !== $this->namedArgumentsResolver) {
+            $arguments = $this->namedArgumentsResolver->resolveArguments($arguments, $class, '__construct');
+        }
+
+        return new $class(...array_values($arguments));
     }
 }
--- a/src/Generator/Instantiator/Chainable/StaticFactoryInstantiator.php
+++ b/src/Generator/Instantiator/Chainable/StaticFactoryInstantiator.php
@@ -16,10 +16,22 @@ namespace Nelmio\Alice\Generator\Instantiator\Chainable;
 use Nelmio\Alice\Definition\MethodCall\NoMethodCall;
 use Nelmio\Alice\Definition\ServiceReference\StaticReference;
 use Nelmio\Alice\FixtureInterface;
+use Nelmio\Alice\Generator\NamedArgumentsResolver;
 use Nelmio\Alice\Throwable\Exception\Generator\Instantiator\InstantiationExceptionFactory;
 
 final class StaticFactoryInstantiator extends AbstractChainableInstantiator
 {
+    /**
+     * @var NamedArgumentsResolver|null
+     */
+    private $namedArgumentsResolver;
+
+    // TODO: make $namedArgumentsResolver non-nullable in 4.0. It is currently nullable only for BC purposes
+    public function __construct(NamedArgumentsResolver $namedArgumentsResolver = null)
+    {
+        $this->namedArgumentsResolver = $namedArgumentsResolver;
+    }
+
     /**
      * @inheritDoc
      */
@@ -45,6 +57,10 @@ final class StaticFactoryInstantiator extends AbstractChainableInstantiator
 
         if (null === $arguments) {
             $arguments = [];
+        }
+
+        if (null !== $this->namedArgumentsResolver) {
+            $arguments = $this->namedArgumentsResolver->resolveArguments($arguments, $factory, $method);
         }
 
         $instance = $factory::$method(...array_values($arguments));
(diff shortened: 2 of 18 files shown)


theofidry:
Nice! Let me know when you think you're done & ready for review :)

julienfalque:
Actually I'd like to have your feedback :)

Especially regarding the new `NamedArgumentsResolver` class:
* is its contract ok?
* should it have an interface?

src/Loader/NativeLoader.php
```diff
@@ -626,7 +629,7 @@ protected function createPropertyAccessor(): PropertyAccessorInterface
 
     protected function createCaller(): CallerInterface
     {
-        return new SimpleCaller($this->getCallProcessor(), $this->getValueResolver());
+        return new SimpleCaller($this->getCallProcessor(), $this->getValueResolver(), new NamedArgumentsResolver());
```
theofidry, line 632: I would add a `createNamedArgumentResolver()` method and add `@method NamedArgumentsResolver get NamedArgumentsResolver()`

src/Generator/Instantiator/Chainable/NoCallerMethodCallInstantiator.php
```diff
@@ -15,9 +15,20 @@
...
+use Nelmio\Alice\Generator\NamedArgumentsResolver;
 
 final class NoCallerMethodCallInstantiator extends AbstractChainableInstantiator
 {
+    /**
+     * @var NamedArgumentsResolver|null
+     */
+    private $namedArgumentsResolver;
+
+    public function __construct(NamedArgumentsResolver $namedArgumentsResolver = null)
```
theofidry, line 27:
```suggestion
    // TODO: make $namedArgumentsResolver non-nullable in 4.0. It is currently nullable only for BC purposes
    public function __construct(NamedArgumentsResolver $namedArgumentsResolver = null)
```

src/Generator/Instantiator/Chainable/StaticFactoryInstantiator.php
```diff
@@ -16,10 +16,21 @@
...
 use Nelmio\Alice\Throwable\Exception\Generator\Instantiator\InstantiationExceptionFactory;
 
 final class StaticFactoryInstantiator extends AbstractChainableInstantiator
 {
+    /**
+     * @var NamedArgumentsResolver|null
+     */
+    private $namedArgumentsResolver;
+
+    public function __construct(NamedArgumentsResolver $namedArgumentsResolver = null)
```
theofidry, line 29: Same TODO here

src/Generator/NamedArgumentsResolver.php
```diff
@@ -0,0 +1,85 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Nelmio\Alice\Generator;
+
+class NamedArgumentsResolver
+{
+    public function resolveArguments(array $arguments, string $className, string $methodName)
```
theofidry, line 18:
```suggestion
    public function resolveArguments(array $arguments, string $className, string $methodName): array
```

?

tests/Generator/NamedArgumentsResolverTest.php
```diff
@@ -0,0 +1,304 @@
...
+            [
+                'baz2' => 'value 2',
+            ],
+            [
+                'value 1',
+                'baz2' => 'value 2',
+            ],
+        ];
+
+        yield 'with method that does not exist' => [
```
theofidry, line 266: AFAICT it's also missing the case where you are passing the numeric keys as strings e.g.`'0'` which is not uncommon if you need a flag e.g. `0 (unique)`

src/Generator/Instantiator/Chainable/NoCallerMethodCallInstantiator.php
julienfalque: Also added in `SimpleCaller`.

mlavrinenko:
@julienfalque excellent work!

I though about using the fork for this feature, but faced a problem in a case like this:

```php
        yield 'with null as passed argument that have nullable value' => [
            DummyWithMethods::class,
            'methodWithNullables',
            [
                'baz1' => 'hello',
                'baz2' => null,
            ],
            [
                'baz1' => 'hello',
                'baz2' => null,
            ],
        ];

    public function methodWithNullables(?string $baz1, ?string $baz2)
    {
        $this->baz1 = $baz1;
        $this->baz2 = $baz2;
    }
```

It throws:
> Nelmio\Alice\Generator\NamedArgumentsResolverTest::testResolveArguments with data set "with null as passed agrument that have default nullable value" ('Nelmio\Alice\Entity\DummyWithMethods', 'methodWithNullables', array('hello', null), array('hello', null))
RuntimeException: Argument baz2 of Nelmio\Alice\Entity\DummyWithMethods::methodWithNullables() is not passed a value and does not define a default one.

Also I would be glad to use it with inheritance ability, but this doesn't work:

```php
namespace Some;
class Entity {
  public function __construct(string $foo, string $bar): {}
}
```
```yaml
Some\Entity:
  entity_template (template):
    __construct:
      foo: 'value 1'
  entity (extends entity_template):
    __construct:
      bar: 'value 2'
```

julienfalque:
@mlavrinenko Good catch! I added the failing test case and fixed it, thanks!

> I would be glad to use it with inheritance ability, but this doesn't work

I'm not familiar with this feature but as far as I understand, it does not support method calls, does it?

mlavrinenko:
> I'm not familiar with this feature but as far as I understand, it does not support method calls, does it?

I suppose so. I just wanted to mention that named parameters support will be extra useful with the ability to inherit a template.
Inheritance for __calls could be ambigous and not possible at all, but for __construct it would be fine.

theofidry:
@julienfalque is there anything left to do/check or is this good to go?

julienfalque:
@theofidry For me the PR is ready for merge unless you think we should work on inheritance as suggested by @mlavrinenko, but I guess that could be a separate PR.

theofidry:
No indeed I think this is another feature. Ok I’ll have a look to merge it
before the next release

julienfalque:
I just pushed a small tweak in error messages: argument names now include the `$` sign, this increases readability IMO.

julienfalque:
Mmh not sure test failures are related :/

tests/Loader/LoaderIntegrationTest.php
```diff
@@ -1978,6 +1978,61 @@ public function provideLegacyFixturesToInstantiate()
             'An error occurred while generating the fixture "dummy" (Nelmio\Alice\Entity\Instantiator\DummyWithNamedPrivateConstructor): Could not instantiate fixture "dummy".',
             GenerationThrowable::class,
         ];
+
+        yield 'with constructor named parameters' => [
```
theofidry, line 1982: I think this is a little error of a merge. The current cases have been appended to `provideLegacyFixturesToInstantiate()` in which the deprecation message is expected (hence the CI failure). They should instead be moved in `provideFixturesToInstantiate()`.

I also notice i did a mistake with a PR earlier and there is still commented cases in `provideFixturesToInstantiate()`. If you can remove them while you at it it would be amazing

julienfalque: Ah thanks for the hint! I overlooked this when rebasing. I removed commented out code too (as a dedicated commit).

theofidry:
Hm looks like one test became flaky under 7.4 :(

julienfalque:
Thanks @theofidry and @mlavrinenko!

theofidry:
Thank you @julienfalque :)

PR #1030, opened 2020-04-30
Fix parsing of numeric, boolean and null values for function arguments

### LINKED ISSUE #1029: Function parameters are always parsed as string even when they are booleans or null
The function arguments since version `3.x` are always being parsed as `string` instead of the right type, e.g.:

```
App\Entity\MyEntity:
    my_entity:
        __construct:
            - '<randomNumber(5, true)>'
```

I would expect `true` to be parsed as `boolean`, but it's not. The same goes for `null`. This breaks all functions with typehinted parameters. It used to work, so to me looks like a regression but I'm not sure about when it has been introduced

ste93cry:
Fixes #1029 by implementing the parsing of `numeric`, `boolean` and `null` values I don't really like that I have to parse two times the value, both in the lexer and in the parser, but as the `Token` class accepts only a `string` as value of the token I don't see any other way around this without widening the typehint. However, as far as I see quite a lot of things are marked as `@internal` so BC should not be a problem if we decide to go that way

### DIFF
--- a/src/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/FunctionTokenParser.php
+++ b/src/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/FunctionTokenParser.php
@@ -143,32 +143,30 @@ final class FunctionTokenParser implements ChainableTokenParserInterface, Parser
         switch (true) {
             case $value === 'true':
                 return true;
+
             case $value === 'false':
                 return false;
+
             case $value === 'null':
                 return null;
-            case ctype_digit($value):
+
+            case preg_match('/^([-+])?([0-9]+)$/', $value, $matches):
                 $castedValue = (int) $value;
 
-                if (0 === $value[0]) {
-                    return octdec($value);
-                } elseif ($value === (string) $castedValue) {
+                if ('0' === $matches[2][0]) {
+                    return '-' === $matches[1] ? -octdec($matches[2]) : octdec($matches[2]);
+                }
+
+                if ($value === (string) $castedValue || ('+' === $matches[1] && $matches[2] === (string) $castedValue)) {
                     return $castedValue;
                 }
 
                 return $value;
-            case $value[0] === '-' && ctype_digit(substr($value, 1)):
-                $castedValue = (int) $value;
 
-                if (0 === $value[0]) {
-                    return -octdec($value);
-                } elseif ($value === (string) $castedValue) {
-                    return $castedValue;
-                }
-
-                return $value;
+            case is_numeric($value):
             case preg_match('/^[-+]?[0-9]*(\.[0-9]+)?$/', $value):
                 return (float) $value;
+
             default:
                 return $parser->parse($value);
         }
--- a/src/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/FunctionTokenParser.php
+++ b/src/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/FunctionTokenParser.php
@@ -113,7 +113,6 @@ final class FunctionTokenParser implements ChainableTokenParserInterface, Parser
         }
 
         $argumentEscaper = $this->argumentEscaper;
-
         $escapedString = preg_replace_callback(
             '/\'(.*?)\'|"(.*?)"/',
             function (array $matches) use ($argumentEscaper): string {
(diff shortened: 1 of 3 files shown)


theofidry:
>However, as far as I see quite a lot of things are marked as @internal so BC should not be a problem if we decide to go that way

Indeed the Expression Language internals are entirely internal, so there is no problem for BC breaks there

ste93cry:
I have encountered a problem and would like to receive advice on how to proceed: at first I though that implementing a lexer was fine, however (I should have though about it before) obviously all values are passed to the lexer regardless of where they came from, be it a YAML property or a function argument. I should limit the lex to the latter case, but I cannot find any good way to do it so probably my solution is not as feasible as I wish. The only way I can think of is to discard the new token type and move the parsing of the scalar values (what's done now in the lexer) into the `FunctionTokenParser` class, along with the special handling of string values. WDYT?

theofidry:
>be it a YAML property or a function argument. I should limit the lex to the latter case,

Indeed, that's definitely a limitation from the current design

>the scalar values (what's done now in the lexer) into the FunctionTokenParser class, along with the special handling of string values.

I think that would be more appropriate: at least you have more context there whereas when being in the lexer, you are just to create tokens and nothing more

ste93cry:
I fixed all tests and build is now green :tada:

src/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/FunctionTokenParser.php
```diff
@@ -129,13 +128,49 @@ function (array $matches) use ($argumentEscaper): string {
...
 
         return $arguments;
     }
+
+    private function parseArgument(ParserInterface $parser, string $value)
+    {
+        switch (true) {
+            case $value === 'true':
+                return true;
+            case $value === 'false':
```
theofidry, line 146: could you put a line return after the `return` and bellow as well?

src/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/FunctionTokenParser.php
```diff
@@ -129,13 +128,49 @@ function (array $matches) use ($argumentEscaper): string {
...
+                return false;
+            case $value === 'null':
+                return null;
+            case ctype_digit($value):
+                $castedValue = (int) $value;
+
+                if (0 === $value[0]) {
+                    return octdec($value);
+                } elseif ($value === (string) $castedValue) {
```
theofidry, line 155: no need for an else if your return before

src/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/FunctionTokenParser.php
```diff
@@ -129,13 +128,49 @@ function (array $matches) use ($argumentEscaper): string {
...
+                    return $castedValue;
+                }
+
+                return $value;
+            case $value[0] === '-' && ctype_digit(substr($value, 1)):
+                $castedValue = (int) $value;
+
+                if (0 === $value[0]) {
+                    return -octdec($value);
+                } elseif ($value === (string) $castedValue) {
```
theofidry, line 165: same here

src/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/FunctionTokenParser.php
ste93cry: There are three different cases to cover: I have to either split them into `if/elseif/else` statements or into multiple `if` statements 🤔 I'm not sure I get what I should change

theofidry: Instead of:

```php
if (x) {
  // ...
  return ...;
} elseif (y) {
  // ...
  return ...;
} else {
  // ...
  return ...;
}

```

A strict equivalent is:

```php
if (x) {
  // ...
  return ...;
}

if (y) {
  // ...
  return ...;
}
  
// ...
return ...;

```

ste93cry:
Fixed all CR issues and a bug that I spotted. Do you have any ETA of the next patch version?

theofidry:
I'll try this weekend

theofidry:
Thank you @ste93cry!

PR #1038, opened 2020-06-03
Fix method name on exception

robertbasic:
Hey, I misconfigured my alice files so I ran into this error where the CallProcessorExceptionFactory called a non-existent method.


theofidry:
Looks like `OptionalValueResolverTest::testCanHandleExtremaQuantifiersCorrectly` is now failing randomly on 7.4 :/

theofidry:
Thank you @robertbasic

PR #1039, opened 2020-06-05
Add WordPress third-party lib

pretzlaw:
WordPress entities work a bit different than common ones.
Sometimes the constructor, getter/setter already access the database,
fields are internally used but missing or some properties are fetched from DB on the fly.
The rmp-up/wp-fixtures enhances the NativeLoader
to bypass such obstacles and allows creating native WP entities.

* Maintained and enhancing for two years already
* Using alice for quiet a time now
* Finally works with native WordPress entities since v0.8


theofidry:
Thanks you @pretzlaw :)

PR #1047, opened 2020-08-09
fix returned empty arguments when args are passed via func_get_args ()

### LINKED ISSUE #1046: 3.7.0 introduce BC issue with __construct and func_get_args()
After upgrading from 3.6 to 3.7 I have issue on fixtures using __construct on [CrEOF\Spatial\PHP\Types\Geography\Point](https://github.com/creof/doctrine2-spatial/blob/58ea5fae1c1b450ee08d7dac25cd9e8f5e6fdebd/lib/CrEOF/Spatial/PHP/Types/Geography/Point.php) objects that could be initialized with `$x`, `$y` (floats coordinates) as constructor  parameters.

Here's the fixture 
```yaml
CrEOF\Spatial\PHP\Types\Geography\Point:
  Point1:
    __construct: [2.328344, 48.870193]
```
And the error is 
```sh
In SimpleObjectGenerator.php line 114:
  An error occurred while generating the fixture "Point1" (CrEOF\Spatial\PHP\Types\Geography\Point): Could not instantiate fixture "Point1".  

In InstantiationExceptionFactory.php line 26:
  Could not instantiate fixture "Point1".  

In AbstractPoint.php line 210:                                                                                
  Invalid parameters passed to CrEOF\Spatial\PHP\Types\Geography\Point::__construct:  
```

I've also tried with the new named arguments feature
```yaml
CrEOF\Spatial\PHP\Types\Geography\Point:
  Point1:
    __construct:
      x: 2.328344
      y: 48.870193
```

And the error
```sh
In SimpleObjectGenerator.php line 114:                                                                        
  An error occurred while generating the fixture "Point1" (CrEOF\Spatial\PHP\Types\Geography\Point): Could not instantiate fixture "Point1".  

In InstantiationExceptionFactory.php line 26:
  Could not instantiate fixture "Point1".  
                                           
In NamedArgumentsResolver.php line 75:                            
  Unknown arguments for CrEOF\Spatial\PHP\Types\Geography\Point::__construct(): $x, $y.  
```

Here's the constructor of the abstract class
```php
public function __construct()
{
    $argv = $this->validateArguments(func_get_args());

    call_user_func_array(array($this, 'construct'), $argv);
}
```

If I call `var_dump(func_get_args())`, in 3.7 the output is an **empty array**, in 3.6 it's an array with fixture's arguments `[2.328344, 48.870193]`

It seems that `__construct` is not working anymore with `func_get_args()`. 

if I update the constructor as below, the issue is fixed.
```php
public function __construct(float $x, float $y)
{
    call_user_func_array(array($this, 'construct'), [$x, $y]);
}
```

Issue comes from that commit c64ba862215c7943fe2c3e150aa91c9169e8244b, exactly from that method https://github.com/nelmio/alice/blob/570a698ed5c9f693f5765e8bcc2599faadbc22c3/src/Generator/NamedArgumentsResolver.php#L18 that return an empty array because it only checks the method parameters arguments. But when we use `func_get_args()`, `$method->getParameters()` returns an empty array.

(ping @julienfalque)

julienbornstein:
- add fix #1046
- add non-regression test

### DIFF
--- a/fixtures/Entity/DummyWithNoArgumentConstructor.php
+++ b/fixtures/Entity/DummyWithNoArgumentConstructor.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of the Alice package.
+ *
+ * (c) Nelmio <hello@nelm.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Nelmio\Alice\Entity;
+
+class DummyWithNoArgumentConstructor
+{
+    /**
+     * @var array
+     */
+    private $args;
+
+    public function __construct()
+    {
+        $this->args = func_get_args();
+    }
+
+    /**
+     * @return array
+     */
+    public function getArgs(): array
+    {
+        return $this->args;
+    }
+}
--- a/src/Generator/NamedArgumentsResolver.php
+++ b/src/Generator/NamedArgumentsResolver.php
@@ -23,6 +23,10 @@ class NamedArgumentsResolver
             return $arguments;
         }
 
+        if (0 === count($method->getParameters())) {
+            return $arguments;
+        }
+
         $sortedArguments = [];
         $buffer = [];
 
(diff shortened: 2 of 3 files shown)


theofidry:
@julienbornstein thanks for the PR! Looks like a dependency needs to be updated to fix the CI, I'll try to have a look at it either tonight or this weekend

julienbornstein:
😴

theofidry:
Thank you @julienbornstein, sorry it went under the radar

Issue #1056, opened 2020-10-29
PHP 8.0 invalid scalar type in NullRangeNameDenormalizer


dotdevru:
On php 8.0 function str_replace called with $currentIndex which is int, but in php 8 it should be exactly string:

```php
str_replace(): Argument #2 ($replace) must be of type array|string, int given

vendor/nelmio/alice/src/FixtureBuilder/Denormalizer/Fixture/Chainable/NullRangeNameDenormalizer.php:81
vendor/nelmio/alice/src/FixtureBuilder/Denormalizer/Fixture/Chainable/CollectionDenormalizerWithTemporaryFixture.php:83
vendor/nelmio/alice/src/FixtureBuilder/Denormalizer/Fixture/Chainable/CollectionDenormalizerWithTemporaryFixture.php:115
vendor/nelmio/alice/src/FixtureBuilder/Denormalizer/Fixture/Chainable/SimpleCollectionDenormalizer.php:104
vendor/nelmio/alice/src/FixtureBuilder/Denormalizer/Fixture/FixtureDenormalizerRegistry.php:60
vendor/nelmio/alice/src/FixtureBuilder/Denormalizer/Fixture/TolerantFixtureDenormalizer.php:43
vendor/nelmio/alice/src/FixtureBuilder/Denormalizer/Fixture/SimpleFixtureBagDenormalizer.php:83
vendor/nelmio/alice/src/FixtureBuilder/Denormalizer/SimpleDenormalizer.php:46
vendor/nelmio/alice/src/FixtureBuilder/SimpleBuilder.php:39
vendor/nelmio/alice/src/Loader/SimpleDataLoader.php:45
vendor/nelmio/alice/src/Loader/SimpleFileLoader.php:47
vendor/nelmio/alice/src/Loader/NativeLoader.php:259
tests/DataForm/Api/FormDocument/FormDocumentApiTest.php:63
vendor/phpunit/phpunit/src/Util/Annotation/DocBlock.php:427
vendor/phpunit/phpunit/src/Util/Annotation/DocBlock.php:284
vendor/phpunit/phpunit/src/Util/Test.php:319
vendor/phpunit/phpunit/src/Framework/TestBuilder.php:74
vendor/phpunit/phpunit/src/Framework/TestSuite.php:875
vendor/phpunit/phpunit/src/Framework/TestSuite.php:235
vendor/phpunit/phpunit/src/Framework/TestSuite.php:365
[...]

PR #1057, opened 2020-10-31
#1056: PHP 8.0 invalid scalar type in NullRangeNameDenormalizer

dotdevru:
Fixes https://github.com/nelmio/alice/issues/1056


theofidry:
Thank you @dotdevru

PR #1059, opened 2020-11-03
fix(deps): replace deprecated `fzaninotto/faker` by `fakerphp/faker`, close #1058

### LINKED ISSUE #1058: Dependency `fzaninotto/faker` is deprecated
Hi :wave: ,

The `fzaninotto/faker` package has been deprecated, see https://github.com/fzaninotto/Faker and https://marmelab.com/blog/2020/10/21/sunsetting-faker.html.

I think it would be nice to replace it by [`fakerphp/faker`](https://github.com/FakerPHP/Faker) .

I don't have so much time right now but I can work on this when possible. :+1:

Kocal:
Close #1058

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
 
     "require": {
         "php": "^7.3",
-        "fzaninotto/faker": "^1.6",
+        "fakerphp/faker": "^1.10",
         "myclabs/deep-copy": "^1.5.2",
         "sebastian/comparator": "^3.0 || ^4.0",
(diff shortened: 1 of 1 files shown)


theofidry:
I take that there is no API changes atm?

Kocal:
I don't think so, the only changes I see from the original package are:
- support for PHP 8
- some bug fixes

See https://github.com/FakerPHP/Faker/releases

theofidry:
Thank you @Kocal

PR #1060, opened 2020-11-09
Missing autowire parameter

Spomky:
With the given configuration, Symfony is not able to instantiate the custom provider as it requires a Generator to be injected.
The Generator objects comes from the constructor of `Faker\Provider\Base`.

This PR makes it clear that `autowire` should be enabled or that the `Faker\Generator` service should be passed as constructor argument.


theofidry:
Thank you @Spomky

PR #1061, opened 2020-11-15
Fix links to Faker

bramceulemans:


theofidry:
Thank you @bramceulemans

PR #1064, opened 2020-12-09
Add support for PHP 8

Yozhef:


Yozhef:
Error in build PHP 8.0:
https://github.com/theofidry/composer-inheritance-plugin/pull/8

simPod:
https://github.com/nelmio/alice/blob/25e127fa286d0a90f62b79683c05eb63ca23a86a/src/Throwable/Exception/FixtureBuilder/Denormalizer/DenormalizerExceptionFactory.php#L33 needs to add `= null` to throwable so required param is not after optional one.

theofidry:
Thanks for the PR @Yozhef, but I think it will require a migration to GitHub Actions as Travis CI kinda gave up on OSS support :(

Yozhef:
@theofidry ok I will do it

Yozhef:
@theofidry 
A little need for your consultation:
Changing the function will cause what needs to be changed or updated for this structure? 

File -> SimpleReferenceTokenParserTest.php:56 -> SimpleReferenceTokenParser -> FixtureReferenceValue
```
In PHP 7.4
var_dump(substr('', 1)); // false

In PHP 8.0
var_dump(substr('', 1)); // ""
```

tests/FixtureBuilder/ExpressionLanguage/TokenTypeTest.php
```diff
@@ -48,7 +48,7 @@ public function testStaticValues(): void
 
         $reflProp = $reflClass->getProperty('values');
         $reflProp->setAccessible(true);
-        $values = $reflProp->getValue(TokenType::class);
+        $values = $reflProp->getValue((object) TokenType::class);
```
theofidry, line 51: ??

theofidry:
@Yozhef sorry I'm not following, what's the issue?

tests/FixtureBuilder/ExpressionLanguage/TokenTypeTest.php
Yozhef: If run in PHP 8
```
1) Nelmio\Alice\FixtureBuilder\ExpressionLanguage\TokenTypeTest::testStaticValues
TypeError: ReflectionProperty::getValue(): Argument #1 ($object) must be of type ?object, string given

/home/runner/work/alice/alice/tests/FixtureBuilder/ExpressionLanguage/TokenTypeTest.php:50
```

theofidry: Maybe the reflection API changed for a static value, but the change does not look correct, `(object) TokenType::class === stdClass{ #scalar: TokenType::class }` which is most likely not what we want

alexislefebvre:
@Yozhef Congrats on making the CI green. :smiley:

alexislefebvre (approved):

src/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/SimpleReferenceTokenParser.php
```diff
@@ -41,7 +41,9 @@ public function parse(Token $token): FixtureReferenceValue
         $value = $token->getValue();
 
         try {
-            return new FixtureReferenceValue(substr($value, 1));
+            $elements = empty($value) ? false : substr($value, 1);
```
theofidry, line 44: Is this necessary? IIRC this parser can only parse `SIMPLE_REFERENCE_TYPE` tokens which have the form `@.+`, i.e. `$value` is a non-empty string here and starts with `@`. Anything else IMO should go with the exception.

Looking at it again though it might be clearer to throw the exception if `$value` is not a string and `$elements` an empty string rather than relying on an exception to be thrown and catch it:

```php

$value = $token->getValue();

if (!is_string($value) || '' === $value) {
    throw ExpressionLanguageExceptionFactory::createForUnparsableToken($token);
}

return new FixtureReferenceValue(substr($value, 1));
```

src/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/VariableTokenParser.php
```diff
@@ -45,7 +45,9 @@ public function canParse(Token $token): bool
      */
     public function parse(Token $token)
     {
-        $variable = substr($token->getValue(), 1);
+        $variable = empty($token->getValue())
```
theofidry, line 48: likewise here; also general note `empty()` should be avoided at all cost because is unreliable: `empty('0') = true`

tests/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/StringArrayTokenParserTest.php
```diff
@@ -71,7 +71,10 @@ public function testThrowsAnErrorIfCouldNotParseToken(): void
             $parser = new StringArrayTokenParser(new FakeParser());
 
             $parser->parse($token);
-            static::fail('Expected exception to be thrown.');
+
+            if (PHP_VERSION_ID < 80000) {
+                static::fail('Expected exception to be thrown.');
```
theofidry, line 76: why is it no longer the case here in PHP8?

theofidry (approved):
👍  looks good to me overall aside a few notes; thanks a lot for the work

tests/FixtureBuilder/ExpressionLanguage/Parser/TokenParser/Chainable/StringArrayTokenParserTest.php
Yozhef: Because PHP8.0 function substr return only string -therefore, the parameter error typehint will never fall.

theofidry:
👌

alexislefebvre:
@theofidry 

> Thanks for the PR @Yozhef, but I think it will require a migration to GitHub Actions as Travis CI kinda gave up on OSS support :(

Did they added some limitations for open source? I found this article and it's not very clear: https://devclass.com/2020/11/25/travis-ci-open-source-engagement/

theofidry:
they change of ownership and the direction is not clear. In the short run it's effectively very limited support and it's unlikely to change. They claim to still be able to offer support for popular OSS projects but I heard they won't decide anything there before another 2-3 months.

TL:DR; Travis is indeed dead for OSS

alexislefebvre:
Ok, thanks for the report.

PR #1065, opened 2020-12-12
Fix class order, when merging data

evolic:
I noticed that ```IncludeDataMerger``` class merges data in wrong way.

For example if I have 3 files as follow:
- file1.yaml
  + Class1a
  + Class1b
- file2.yaml
  + Class2a
  + Class2b
- file3.yaml
  + Class3a
  + Class3b

I noticed that ```mergeInclude()``` method from mentioned class would make following list of objects:
- Class3a
- Class3b
- Class2a
- Class2b
- Class1a
- Class1b

which is wrong and unacceptable for me.

In my case I need to have possibility to make an ordered list:
- create users at first
- create other objects, which relay on the users (user ids)

I fixed ```IncludeDataMerger``` class and related unit tests.

### DIFF
--- a/src/Parser/IncludeProcessor/IncludeDataMerger.php
+++ b/src/Parser/IncludeProcessor/IncludeDataMerger.php
@@ -28,18 +28,21 @@ final class IncludeDataMerger
      */
     public function mergeInclude(array $data, array $includeData): array
     {
-        foreach ($data as $class => $fixtures) {
+        foreach ($includeData as $class => $fixtures) {
             // $class is either a FQCN or 'parameters'
-            $includeData[$class] = (
-                array_key_exists($class, $includeData)
-                    && is_array($includeData[$class])
-                    && is_array($fixtures)
-            )
-                ? array_merge($includeData[$class], $fixtures)
-                : $fixtures
-            ;
+            if (array_key_exists($class, $data)) {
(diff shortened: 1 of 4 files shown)


theofidry:
I'm afraid this cannot be accepted :/

The flow is indeed the expected one, if you have an include, you expect the include to be precessed _first_. This is a very common behaviour (GitLab CI, Symfony config files, etc.).

Besides your and my opinion on this, this is a BC break. It is [the current "inverted order"] also the behaviour in 2.x which was kept in 3.x to avoid BC breaks with no easy migration path.

evolic:
@theofidry I understand your point of view. In my case my changes allow me to create list of components (modules), which will be loaded in certain order (Symfony's ```bundle.php```) and the same order will be used to load all fixtures across the whole application.

I don't use ```include``` keyword to include many files inside main one:
e.g. in https://github.com/ergonode/backend/blob/master/src/Resources/fixtures/develop/fixture.yaml

I followed the other approach to do the same thing.

BTW. The library is awesome and it helps me a lot with creating test data.

theofidry:
@evolic I really appreciated the patch but as explained in my previous command, this unfortunately cannot be merged. I am not entirely too sure but maybe it would be possible to override the IncludeDataMerger somehow for your specific case

Issue #1035, opened 2020-05-22
NamedArgumentsResolver throws exception for varidic functions


ndench:
Given this entity:

```php
class Message
{
    // ...

    public function __construct(string $body, string $email, File ...$attachments)
    {
        $this->body         = $body;
        $this->email        = $email;
        $this->attachments  = new ArrayCollection($attachments);
    }

```

And this fixture:

```yaml
App\Entity\Message:
    message:
        __construct:
            - <sentence()>
            - <email()>
```

I receive the following error:
```
RuntimeException: Argument $attachments of App\Entity\Message::__construct() is not passed a value and does not define a default one.
```

<details>
<summary>Full stack trace</summary>

```
Nelmio\Alice\Throwable\Exception\Generator\DebugUnexpectedValueException: An error occurred while generating the fixture "message" (App\Entity\Message): Could not instantiate fixture "message".
vendor/nelmio/alice/src/Generator/ObjectGenerator/SimpleObjectGenerator.php:114
vendor/nelmio/alice/src/Generator/ObjectGenerator/CompleteObjectGenerator.php:53
vendor/nelmio/alice/src/Generator/DoublePassGenerator.php:61
vendor/nelmio/alice/src/Generator/DoublePassGenerator.php:49
vendor/nelmio/alice/src/Loader/SimpleDataLoader.php:49
vendor/nelmio/alice/src/Loader/SimpleFilesLoader.php:64
vendor/theofidry/alice-data-fixtures/src/Loader/SimpleLoader.php:53
vendor/theofidry/alice-data-fixtures/src/Loader/PersisterLoader.php:83
vendor/theofidry/alice-data-fixtures/src/Loader/PurgerLoader.php:119
vendor/liip/test-fixtures-bundle/src/Services/DatabaseTools/AbstractDatabaseTool.php:153
vendor/liip/test-fixtures-bundle/src/Test/FixturesTrait.php:101
tests/Integration/Repository/MessageRepository/FindByDealTest.php:26

Caused by
Nelmio\Alice\Throwable\Exception\Generator\Instantiator\InstantiationException: Could not instantiate fixture "message".
vendor/nelmio/alice/src/Throwable/Exception/Generator/Instantiator/InstantiationExceptionFactory.php:26
vendor/nelmio/alice/src/Generator/Instantiator/Chainable/AbstractChainableInstantiator.php:48
vendor/nelmio/alice/src/Generator/Instantiator/InstantiatorRegistry.php:72
[...]

PR #1077, opened 2021-02-12
Fixes: #1035

jvmanji:
See discussion in: https://github.com/nelmio/alice/issues/1035

### DIFF
--- a/fixtures/Entity/DummyWithVariadicConstructorParamAndParam.php
+++ b/fixtures/Entity/DummyWithVariadicConstructorParamAndParam.php
@@ -1,0 +1,26 @@
+<?php
+
+/*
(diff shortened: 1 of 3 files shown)


theofidry:
👍 would it be possible to have a test case as well to avoid regressions in the future?

jvmanji:
I've added test case in `tests/Generator/NamedArgumentsResolverTest.php`, because I think it belongs there. Can you check whether naming of things suits you? Feel free to send any remarks.

kotbehemot53:
Can anyone merge it? It would be really nice to have the fix in the base branch.

theofidry:
Thank you @jvmanji

Issue #1080, opened 2021-03-11
Fixture wildcard reference is order dependent


ndench:
When loading fixtures into a development database, fixtures are instantiated in different orders on different development machines. This becomes an issue when we have fixtures dependent on one and other:

```yaml
App\Entity\User:
    user_{@company.*}:
        company: <current()>
```

Stepping through the instantiation I can see that \Nelmio\Alice\FixtureBuilder\Denormalizer\Fixture\Chainable\ReferenceRangeNameDenormalizer::buildReferencedValues tries to find built fixtures that match company.*. For some development machines, this works correctly. However for others, the Company is instantiated later which means there are no built fixtures that match at this point. Thus the method returns an empty array and the User fixture is not instantiated at all, but everything else continues as normal.

Initially I thought this issue was related to #1040 but it's actually a different issue specific for wildcard references.

Proposed solutions:
1. Ensure that wildcard references always return [at least one built fixture](https://github.com/nelmio/alice/blob/f174420601c1115e1a16cdee8320b1c7176ae2d2/src/FixtureBuilder/Denormalizer/Fixture/Chainable/ReferenceRangeNameDenormalizer.php#L130), by throwing FixtureNotFoundException and retrying the fixture
2. Denormalize wildcard references before the [FixtureBagDenormalizer iterates the fixtures](https://github.com/nelmio/alice/blob/f174420601c1115e1a16cdee8320b1c7176ae2d2/src/FixtureBuilder/Denormalizer/Fixture/SimpleFixtureBagDenormalizer.php#L77), the wildcard needs to be removed and new fixtures with references that do not contain wildcards. This will cause the [ReferenceRangeNameDenormalizer to look for a specific built fixture](https://github.com/nelmio/alice/blob/f174420601c1115e1a16cdee8320b1c7176ae2d2/src/FixtureBuilder/Denormalizer/Fixture/Chainable/ReferenceRangeNameDenormalizer.php#L117) and throw if it cannot be found.

PR #1081, opened 2021-03-11
Fix reference range wildcard order

ndench:
I decided to go with solution 1 in #1080 as it seems much more simple and fits nicely with the existing architecture, as opposed to adding in a extra denormalization step to break up wildcard references.

Before this change, wildcard references which are processed before their dependencies are ignored and not created. This change ensures that at least one reference is found that matches the wildcard (and forces a retry if none exist). 

This causes a potential BC break in two cases:
1. You were relying on defined fixtures not being created because they are processed too early. This is likely not what you intended as you have specified the fixture, you probably want it to be created. With this change, it will now be created. 
    ```yaml
    Nelmio\Alice\Entity\User:
        # This will potentially be skipped before because userDetails1 & userDetails2 may not be created yet
        user_{@userDetails*}:
            details: <current()>

    Nelmio\Alice\Entity\UserDetails:
        userDetails1: ~
        userDetails2: ~
    ````
2. You have multiple fixtures which match the wildcard, but don't want some to be matched. This is arguably an incorrect fixture definition, you should ensure that your wildcard only matches the fixtures you want it to match.
    ```yaml
    Nelmio\Alice\Entity\User:
        # This will potentially be called with userDetails_files now, depending on the order of execution
        user_{@userDetails*}:
            details: <current()>

    Nelmio\Alice\Entity\UserDetails:
        userDetails1: ~
        userDetails2: ~

    Nelmio\Alice\Entity\UserDetailsFiles:
        userDetails_files: ~
    ````

I added two tests:
1. Happy path, when referenced fixtures already exist and can be retrieved from the FixtureBag
2. When referenced fixtures do not exist yet, an exception should be thrown, which will cause the [FixtureBagDenormalizer to retry the fixture later](https://github.com/nelmio/alice/blob/f174420601c1115e1a16cdee8320b1c7176ae2d2/src/FixtureBuilder/Denormalizer/Fixture/SimpleFixtureBagDenormalizer.php#L92).

Closes: #1080

### DIFF
--- a/src/FixtureBuilder/Denormalizer/Fixture/Chainable/ReferenceRangeNameDenormalizer.php
+++ b/src/FixtureBuilder/Denormalizer/Fixture/Chainable/ReferenceRangeNameDenormalizer.php
@@ -28,6 +28,7 @@ use Nelmio\Alice\FixtureBuilder\Denormalizer\FlagParserInterface;
 use Nelmio\Alice\FixtureInterface;
 use Nelmio\Alice\IsAServiceTrait;
 use Nelmio\Alice\Throwable\Exception\FixtureBuilder\Denormalizer\FlagParser\FlagParserExceptionFactory;
+use Nelmio\Alice\Throwable\Exception\FixtureNotFoundExceptionFactory;
 use Nelmio\Alice\Throwable\Exception\LogicExceptionFactory;
 
 final class ReferenceRangeNameDenormalizer implements ChainableFixtureDenormalizerInterface, FlagParserAwareInterface
@@ -51,17 +52,17 @@ final class ReferenceRangeNameDenormalizer implements ChainableFixtureDenormaliz
         $this->specsDenormalizer = $specsDenormalizer;
         $this->flagParser = $parser;
     }
-    
+
     public function withFlagParser(FlagParserInterface $parser): self
     {
         return new self($this->specsDenormalizer, $parser);
     }
-    
+
     public function canDenormalize(string $name, array &$matches = []): bool
     {
         return 1 === preg_match(self::REGEX, $name, $matches);
     }
-    
+
     public function denormalize(
         FixtureBag $builtFixtures,
         string $className,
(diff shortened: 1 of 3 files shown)


ndench:
Failed builds appear to be because coverage data is not generated on PHP 7.4. I'm unsure why though.

theofidry:
Let's see if the people reporting the issue can try it out. I am a bit worried about the BC side but it's fair to try this before a more complex solution

ndench:
I've been using this fork in my team of 5 devs since I opened this PR (coming up on 6 months now) and haven't noticed any issues.
Maybe you could try installing it @pierre-H and let us know if you have any issues?

theofidry:
Checking this again it looks good to me. I'm wondering though would there be a way to reproduce it via the IntegrationLoaderTest?

ndench:
Do you mean the [LoaderIntegrationTest](https://github.com/nelmio/alice/blob/5fe627cdfbacba796280093e48e46a7a97c4e6c2/tests/Loader/LoaderIntegrationTest.php)? I think it could probably be introduced there. What exactly would you like out of tests there that is not given by the ReferenceRangeDenormalizerTests that I added?

theofidry:
>Do you mean the LoaderIntegrationTest?

yes

>What exactly would you like out of tests there that is not given by the ReferenceRangeDenormalizerTests that I added?

The tests added to `ReferenceRangeDenormalizerTests` are unit tests are help to cover the newly added path, but to test the feature itself having an integration test via `LoaderIntegrationTest` is nice to have too.

ndench:
Ah yes, that makes sense. I'll add the test in when I get a chance, in the next week or so.

ndench:
Added a LoaderIntegrationTest. Commenting out my change to ReferenceRangeNameDenormalizer (line 132) causes this new test to fail, because now UserDetail fixtures are instantiated.

theofidry:
Thanks a lot @ndench!

PR #1082, opened 2021-03-29
MEzzio/Laminas bridge

func0der:
This is a bridge for nelmio/alice in Laminas.

Maybe the bridges should lay outside of this repo completely?
Generally this should help laminas developers a lot.


func0der:
How is the status about this, @theofidry ?

theofidry:
@func0der sorry for the late reply. The code looks good to me. I'm however slightly uncomfortable to merge this without any test whatsoever, would it be possible to have an equivalent to the Symfony `LoaderIntegrationTest` one? It's not a requirement, but without it as I'm not a Laminas user I won't be able to guarantee if anything breaks for it.

>Maybe the bridges should lay outside of this repo completely?
Generally this should help laminas developers a lot.

Yes and no: in the end it's yet another repository to manage and it's quite annoying to realise you are breaking something for a peculiar bridge and need to come back to fix something in the library for it. In this case, the framework bridges are also very thin layers: they simply register services to the framework DI. Unlike HautelookAliceBundle for example which relies a lot more on the Symfony API to provide more Symfony specific features.

theofidry:
Revisiting this, I think there is just not enough interest for this bridge hence I'll close it down for now. If there is someone interested in it this can be revisited.

func0der:
@lkck24 FYI

PR #1090, opened 2021-09-11
Add PHP 8.1 support

javer:

PR #1091, opened 2021-09-26
Add PHP 8.1 support 

theofidry:
Replacement for https://github.com/nelmio/alice/pull/1090


theofidry:
Note to myself: not to squash PRs where there is multiple authors as it appears to remove them

PR #1093, opened 2021-10-12
Bump dependencies

theofidry:
- Remove support for PHP 7.3
- Remove support for Symfony 2.8, 3.4, 4.x except 4.4 (the LTS), 5.x except 5.3 (current stable)


theofidry:
@chalasr ah right. Well actually the code bump is to 5.3 so looks like I only messed up the issue description!

PR #1097, opened 2021-11-27
Suppress deprecations found in my appl.

mdeherder:


jmsche:
Hi,

I think this PR can be closed as #1099 has fixed it already.

PR #1099, opened 2021-12-06
Add Symfony 6.0 support

javer:
Adds Symfony 6.0 support and fixes tests with Symfony 5.4.

Please note that it requires PHP version bump to 8.0, because `symfony/property-access` since 6.0 has declared return types, and we have to declare them too, and one of them is `mixed` which is supported only since PHP 8.0.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,20 +20,24 @@
     ],
 
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php": "^8.0",
         "fakerphp/faker": "^1.10",
         "myclabs/deep-copy": "^1.10",
         "sebastian/comparator": "^3.0 || ^4.0",
-        "symfony/property-access": "^4.4 || ^5.2",
-        "symfony/yaml": "^4.4 || ^5.2"
+        "symfony/property-access": "^4.4 || ^5.4 || ^6.0",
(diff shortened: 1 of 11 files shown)


jmsche:
PR looks okay.

@theofidry Could you please approve the awaiting workflows? Thanks!

composer.json
```diff
@@ -20,20 +20,24 @@
...
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php": "^8.0",
         "fakerphp/faker": "^1.10",
         "myclabs/deep-copy": "^1.10",
         "sebastian/comparator": "^3.0 || ^4.0",
-        "symfony/property-access": "^4.4 || ^5.2",
-        "symfony/yaml": "^4.4 || ^5.2"
+        "symfony/property-access": "^4.4 || ^5.2 || ^6.0",
+        "symfony/yaml": "^4.4 || ^5.2 || ^6.0"
```
theofidry, line 28: let's bump it to `^5.4` everywhere too

theofidry (approved):
Looks good to me! I don't think the comments _need_ to be addressed here and now so if you wish to it can be done later

composer.json
javer: @theofidry Done

theofidry:
Thank you @javer!

PR #1115, opened 2022-07-14
Silence deprecation warning on `ObjectBag::getIterator` method

hhamon:
On PHP 8.1.7, Sentry captures the following deprecation error with the latest `nelmio/alice` version:

> Deprecated: Return type of Nelmio\Alice\ObjectBag::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice

### DIFF
--- a/src/ObjectBag.php
+++ b/src/ObjectBag.php
@@ -123,7 +123,8 @@ final class ObjectBag implements IteratorAggregate, Countable
     {
         return count($this->objects);
     }
-    
+
+    #[\ReturnTypeWillChange]
     public function getIterator(): ArrayIterator
     {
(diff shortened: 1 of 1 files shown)


src/ObjectBag.php
```diff
@@ -123,7 +123,8 @@ public function count(): int
     {
         return count($this->objects);
     }
-    
+
+    #[\ReturnTypeWillChange]
     public function getIterator(): ArrayIterator
```
theofidry, line 128: Wouldn't it better to replace the return type by `: Traversable` instead?

theofidry:
Closed via #1123

PR #1118, opened 2022-09-29
Add support for enum in symfony property accessor hydrator

akalineskou:
We're using enums for some properties, and for the fixtures to work we had to do a workaround because it was trying to set a string.

The workaround was basically to convert the enum value to an enum (which we assume is correct, else it should fail)
```php
    public function setDummyEnum(DummyEnum|string $dummyEnum): void
    {
        if (is_string($dummyEnum)) {
            // alice fixture
            $dummyEnum = DummyEnum::from($dummyEnum);
        }

        $this->dummyEnum = $dummyEnum;
    }
```

Today I thought enough was enough, we needed a better solution since they were becoming more and more.

I'm not sure if that's where the fix should go, but that's where the exception was happening, so it was a good starting place.

Tested in a big project with no issues.

### DIFF
--- a/src/Generator/Hydrator/Property/SymfonyPropertyAccessorHydrator.php
+++ b/src/Generator/Hydrator/Property/SymfonyPropertyAccessorHydrator.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Nelmio\Alice\Generator\Hydrator\Property;
 
+use function enum_exists;
 use Nelmio\Alice\Definition\Object\SimpleObject;
 use Nelmio\Alice\Definition\Property;
 use Nelmio\Alice\Generator\GenerationContext;
@@ -24,6 +25,10 @@ use Nelmio\Alice\Throwable\Exception\Generator\Hydrator\HydrationExceptionFactor
 use Nelmio\Alice\Throwable\Exception\Generator\Hydrator\InaccessiblePropertyException;
 use Nelmio\Alice\Throwable\Exception\Generator\Hydrator\InvalidArgumentException;
 use Nelmio\Alice\Throwable\Exception\Generator\Hydrator\NoSuchPropertyException;
+use ReflectionEnum;
+use ReflectionException;
+use ReflectionProperty;
+use ReflectionType;
 use Symfony\Component\PropertyAccess\Exception\AccessException as SymfonyAccessException;
 use Symfony\Component\PropertyAccess\Exception\ExceptionInterface as SymfonyPropertyAccessException;
 use Symfony\Component\PropertyAccess\Exception\InvalidArgumentException as SymfonyInvalidArgumentException;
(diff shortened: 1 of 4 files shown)


src/Generator/Hydrator/Property/SymfonyPropertyAccessorHydrator.php
```diff
@@ -55,6 +58,23 @@ public function hydrate(ObjectInterface $object, Property $property, GenerationC
     {
         $instance = $object->getInstance();
 
+        // check if property is an enum (might not have a type)
+        try {
+            $reflectionType = (new ReflectionProperty($instance, $property->getName()))->getType();
+
+            if (null !== $reflectionType && enum_exists($reflectionType->getName())) {
```
theofidry, line 65: I would rather do this:

- do this block as a fallback (i.e. only in case of failure); it is reflection heavy and we are doing it on a property basis, so if we can skip it it's best to do so. In case of failure, you can then throw the original error
- for readability I would split the block in two methods: `isEnumValue()` and `castValueToEnum()`

theofidry, line 65: not a big deal but I personally like to import the functions used

theofidry (commented):
👍 LGTM; just a few nitpicks but the core of the code is good IMO

src/Generator/Hydrator/Property/SymfonyPropertyAccessorHydrator.php
akalineskou: 👍 

Named it `isEnumProperty` since it made more sense (it checks the property not it's value)

Not 100% sure this is what you meant with those functions.

src/Generator/Hydrator/Property/SymfonyPropertyAccessorHydrator.php
```diff
@@ -71,4 +84,37 @@ public function hydrate(ObjectInterface $object, Property $property, GenerationC
 
         return new SimpleObject($object->getId(), $instance);
     }
+
+    private function isEnumProperty($instance, Property $property): ?ReflectionType
```
theofidry, line 88:
```suggestion
    private function getEnumType($instance, Property $property): ?ReflectionType
```

theofidry, line 88: this could be made static

src/Generator/Hydrator/Property/SymfonyPropertyAccessorHydrator.php
```diff
@@ -71,4 +84,37 @@ public function hydrate(ObjectInterface $object, Property $property, GenerationC
...
+
+        if (!enum_exists($reflectionType->getName())) {
+            // might not be an enum
+            return null;
+        }
+
+        return $reflectionType;
+    }
+
+    private function castValueToEnum(ReflectionType $reflectionType, Property $property): ?Property
```
theofidry, line 110: this could be made static

theofidry (approved):

PR #1124, opened 2022-11-08
Add CompilerPass to check if framework_bundle.property_access is enabled

xElysioN:
This allows to have a more user-friendly error when the `framework_bundle.property_access` is disabled.

### DIFF
--- a/src/Bridge/Symfony/DependencyInjection/Compiler/ConfigureDependencyPass.php
+++ b/src/Bridge/Symfony/DependencyInjection/Compiler/ConfigureDependencyPass.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the Alice package.
+ *
+ * (c) Nelmio <hello@nelm.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Nelmio\Alice\Bridge\Symfony\DependencyInjection\Compiler;
+
+use RuntimeException;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class ConfigureDependencyPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        if (!$container->has('property_accessor')) {
+            throw new RuntimeException('NelmioAliceBundle requires framework_bundle.property_access to be enabled.');
+        }
+    }
+}
--- a/fixtures/Bridge/Symfony/Application/config_property_access_disabled.yml
+++ b/fixtures/Bridge/Symfony/Application/config_property_access_disabled.yml
@@ -1,0 +1,23 @@
+#
+# This file is part of the Alice package.
+#
+# (c) Nelmio <hello@nelm.io>
+#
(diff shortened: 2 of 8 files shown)


src/Bridge/Symfony/DependencyInjection/Compiler/ConfigureDependencyPass.php
```diff
@@ -0,0 +1,28 @@
...
+
+declare(strict_types=1);
+
+namespace Nelmio\Alice\Bridge\Symfony\DependencyInjection\Compiler;
+
+use RuntimeException;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class ConfigureDependencyPass implements CompilerPassInterface
```
theofidry, line 20: Maybe `CheckBundleDependenciesPass` would be more appropriate?

src/Bridge/Symfony/DependencyInjection/Compiler/ConfigureDependencyPass.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    public function process(ContainerBuilder $container): void
+    {
+        if (!$container->has('property_accessor')) {
+            throw new RuntimeException('NelmioAliceBundle requires framework_bundle.property_access to be enabled.');
```
theofidry, line 25: I would elevate this to a `LogicException` or `DomainException`: `RuntimeException` to me implies it's an exception that makes sense to catch and retry at runtime, whereas LogicException & Error is more akin to a compilation error and requires a manual fix (which is the case here).

src/Bridge/Symfony/DependencyInjection/NelmioAliceExtension.php
```diff
@@ -17,17 +17,29 @@
...
-final class NelmioAliceExtension extends Extension
+final class NelmioAliceExtension extends Extension implements PrependExtensionInterface
 {
     const SERVICES_DIR = __DIR__.'/../Resources/config';
-    
+
+    public function prepend(ContainerBuilder $container): void
+    {
+        if (isset($container->getExtensions()['framework'])) {
+            $container->prependExtensionConfig('framework', [
```
theofidry, line 35: I do not understand that piece, doesn't that mean the bundle _enables_ the property access? I don't think it's a good idea for the bundle to attempt to change the default settings like so. Your error is a better option IMO, it shouldn't be hard for the user to figure it out and would avoid any surprises of having the property access component enabled when though disabled

theofidry (approved):
Nice! I completely forgot about that setting and that explains a few issues that popped up over time. Thanks a lot :)

src/Bridge/Symfony/DependencyInjection/NelmioAliceExtension.php
xElysioN: Yes with that, the bundle will enable the property_access. 
It was just to avoid having an error when you require the package.
I'll remove this part then :+1:

src/Bridge/Symfony/DependencyInjection/Compiler/ConfigureDependencyPass.php
xElysioN: Yeah, it's better, I was not inspired when I choose this name :D

theofidry:
Thank you @xElysioN!

PR #1127, opened 2023-02-03
Add support for sebastian/comparator 5

louismariegaborit:
With phpunit/phpunit 10.0 release, sebastian/comparator was updated to 5.0.0.  
To migrate project with this new version, alice package should support this new version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "php": "^8.1",
         "fakerphp/faker": "^1.10",
         "myclabs/deep-copy": "^1.10",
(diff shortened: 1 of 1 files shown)


curry684:
Would be great to merge this, currently only blocker on PHPunit 10 install in all my projects.

curry684:
Thanks @theofidry   😄 

For those wondering - now you can upgrade to phpunit 10 but let's just say the current `symfony/phpunit-bridge` should have a conflict with phpunit 10 as it is completely broken hehe.

Issue #1116, opened 2022-07-30
Can't access nested reference property


NotionCommotion:
The first variant cannot indeed work due to conflicting with the existing features.

The second however should work, but maybe `@self->city->postalCodes` will do?

_Originally posted by @theofidry in https://github.com/nelmio/alice/issues/1019#issuecomment-678705589_

Issue #1019, opened 2020-02-27
How to access nested reference property?


StanislavUngr:
Hello guys. So I have this issue where i need to access 2nd level reference's property. How do I do this?

This is my schema:

```
Entity\City:
   postalCodes: string[];
```

```
Entity\Address:
   city: Entity\City
   postalCode: random of (Entity\City->postalCodes)
```

YAML for alice:

```
App\Entity\City\City:
  city_{1..100}:
    __construct:
      name: <city()>
      district: '@district_*'
    postalCodes (unique): 5x <postcode()>
```

```
App\Entity\Address\Address:
  address_{1..100}:
    __construct:
      city: '@city_*'
      postalCode: '<randomElement(@self->postalCodes)>'
      type: <randomEnum(AddressType)>
```

However i get errors everytime i try to invoke <randomElement(**@self->postalCodes**)>:
```
[Nelmio\Alice\Throwable\Exception\ObjectNotFoundException]
  Could not find the object "address_1" of the class "App\Entity\Address\Address".
```
 or <randomElement(**$city->postalCodes**)>:
```
[Nelmio\Alice\Throwable\Exception\Generator\Resolver\UnresolvableValueException]
  Could not evaluate the expression "$city".
```

It seems only 1st level of nesting is allowed.

Do you have any fixes/workarounds how to solve this? Thanks in advance..

lugus:
Hello,

Like you have 100 city and 100 address, maybe you could try this:

```
App\Entity\Address\Address:
  address_{1..100}:
    __construct:
      city: '@city_<current()>'
      postalCode: '<randomElement(@city_<current()>->postalCodes)>'
      type: <randomEnum(AddressType)>
```

StanislavUngr:
Hello,

although this is a temporary fix and destroys logic of M:N relation, for our testing purposes its enough atm.

Thanks for your help 👍 

_PS: PR to implement this feature is welcomed, of course 😄_

J3m5:
The  `@self->city->postalCodes` syntax doesn't works unfortunately, is there another way to access properties through multiple relations ?

 ```
[...]

PR #1129, opened 2023-02-08
Add doc to explain how to access nested property from relation.

WedgeSama:
I thinks the right syntaxe must be something like `@self->created_by->info->created`, but like said it #1116 and refered in #1019, it does not work for now. Using this syntax `@self->created_by.info.created`, it works.

Do you think I have to add in the doc that it is a workaround for now?


theofidry:
Thank you @WedgeSama, note though that this will not work if the property access component is not used/enabled.

PR #1145, opened 2023-11-07
Support Symfony 7

endroid:
Adds support for Symfony 7

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,20 +24,20 @@
         "fakerphp/faker": "^1.10",
         "myclabs/deep-copy": "^1.10",
         "sebastian/comparator": "^3.0 || ^4.0 || ^5.0",
-        "symfony/property-access": "^5.4 || ^6.0",
-        "symfony/yaml": "^5.4 || ^6.0"
+        "symfony/property-access": "^5.4 || ^6.0 || ^7.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -24,20 +24,20 @@
         "fakerphp/faker": "^1.10",
         "myclabs/deep-copy": "^1.10",
         "sebastian/comparator": "^3.0 || ^4.0 || ^5.0",
-        "symfony/property-access": "^5.4 || ^6.0",
-        "symfony/yaml": "^5.4 || ^6.0"
+        "symfony/property-access": "^5.4 || ^6.0 || ^7.0",
```
alexislefebvre, line 27: What about changing `^6.0` to `^6.4` in order to only support the LTS versions and the last release?

It may be done on another PR though.

endroid: @theofidry What do you prefer?

theofidry: I think it's a good idea, but definitely in a different PR. I think it would also make sense to drop 5.4, maintaining 3 different versions is too much IMO

theofidry (commented):
Thanks for the PR! The other changes can be done separately, but I think it would be good to have at least one job where Symfony 7 is forced: https://github.com/nelmio/alice/blob/master/.github/workflows/ci.yaml#L23

theofidry:
Thank you @endroid, merged via #1156

PR #1160, opened 2023-12-16
fix: Improve the error message upon unresolvable value

Rafikooo:
Before:
image
After:
image


src/Throwable/Exception/Generator/Resolver/UnresolvableValueDuringGenerationExceptionFactory.php
```diff
@@ -23,7 +23,7 @@ final class UnresolvableValueDuringGenerationExceptionFactory
     public static function createFromResolutionThrowable(ResolutionThrowable $previous): UnresolvableValueDuringGenerationException
     {
         return new UnresolvableValueDuringGenerationException(
-            'Could not resolve value during the generation process.',
+            $previous->getMessage(),
```
theofidry, line 26: What about: `'Could not resolve value during the generation process: '.$previous->getMessage()`?

theofidry:
Merged via #1188

PR #1204, opened 2024-07-03
Remove usage of the deprecated base extension from HttpKernel

stof:
The base extension available in HttpKernel only adds support for registering some classes to be added in the cache warmer of the doctrine/annotations parser, compared to the base class of the DI component. As annotations are not supported anymore in Symfony 7, this base class has been deprecated in Symfony 7.1.
As this bundle does not rely on the extra feature, it can use the base class from DI directly.


theofidry:
Thanks

PR #1208, opened 2024-08-02
Explicit Error Message For Missing Properties During Hydration

Rafikooo:
Before:
image
After:
image


theofidry:
Thank you @Rafikooo! But I wonder with the regular output don't you get a mention of the original exception?

Rafikooo:
Hey @theofidry,
Could you clarify what you exactly mean?

theofidry:
There should be a previous exception from this `DebugUnexpectedValueException`, I find curious it does not appear there

Issue #1230, opened 2024-12-19
Nondeterministic optional method calls


brainlock:
Hi,

I stumbled upon nondeterministic behavior in the way optional method calls are generated:

```php
<?php

require_once __DIR__.'/vendor/autoload.php';

class Foo {
    public function __construct(
        public string $a = "",
        public int $b = 0,
        public int $c = 0,
        public int $d = 0,
    ) {}

    public function setC(int $c): void {
        $this->c = $c;
    }

    public function setD(int $d): void {
        $this->d = $d;
    }
}

$loader = new Nelmio\Alice\Loader\NativeLoader();
$objectSet = $loader->loadData([
    Foo::class => [
        'foo' => [
            'a' => '<word()>',
            'b' => '<numberBetween(0, 100)>',
            '__calls' => [
                ['setC' => [ 42 ]],
                ['setD (50%?)' => [ 42 ]],
            ],
        ],
    ],
]);

print_r($objectSet->getObjects());
```

If we invoke the script a few times, we see that $a, $b and $c have stable pseudorandom values, while $d
flip-flops between 0 and 42.

```
% for i in $(seq 1 5); do php repro.php; done
Array
(
    [foo] => Foo Object
        (
            [a] => consequatur
            [b] => 25
            [c] => 42
            [d] => 42
        )

)
Array
(
    [foo] => Foo Object
        (
            [a] => consequatur
            [b] => 25
            [c] => 42
            [d] => 42
        )

)
Array
(
    [foo] => Foo Object
        (
            [a] => consequatur
            [b] => 25
            [c] => 42
            [d] => 0
        )

)
Array
(
    [foo] => Foo Object
        (
            [a] => consequatur
            [b] => 25
            [c] => 42
            [d] => 0
        )

)
Array
(
    [foo] => Foo Object
        (
            [a] => consequatur
            [b] => 25
            [c] => 42
            [d] => 42
        )

)
```

The immediate cause for this is that OptionalMethodCallProcessor was changed from using mt_rand to random_int here in this commit: https://github.com/nelmio/alice/commit/a7e15b08a64d4f6fb64b4c3b58bd6a710f71050e#diff-08609f7290422a7652a1827996b95b93293221086ef456fa539bf0770c4b117bR68

Arguably though it shouldn't have been using mt_rand() directly in the first place, but rather use the FakerGenerator instance like all other value generators in Alice. (Then Faker uses mt_rand internally, which is why it happened to work before the switch to random_int, but that's another story)

[...]

PR #1231, opened 2024-12-19
fix nondeterminism in OptionalMethodCallProcessor

brainlock:
An optional method call (`__calls : [ setLocation (80%?): [40.689269, -74.044737] ]`) would be resolved using an external source of randomness, rather than the FakerGenerator instance used by all other value generators, resulting in nondeterministic behavior.

We're making the generator an optional parameter to OptionalMethodCallProcessor and falling back to random_int for backwards-compatibility. This can be made non nullable in a next major release.

(Also see OptionalValueResolver, this is handled in the same way there https://github.com/nelmio/alice/blob/a7e15b08a64d4f6fb64b4c3b58bd6a710f71050e/src/Generator/Resolver/Value/Chainable/OptionalValueResolver.php#L101-L103)

Fixes gh-1230

### DIFF
--- a/src/Generator/Caller/Chainable/OptionalMethodCallProcessor.php
+++ b/src/Generator/Caller/Chainable/OptionalMethodCallProcessor.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Nelmio\Alice\Generator\Caller\Chainable;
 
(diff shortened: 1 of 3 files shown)


theofidry:
Thank you @brainlock!

PR #1284, opened 2025-11-07
Add Symfony8 support

acrobat:
I've also updated the symfony version constraints to only allow support symfony 7 versions and switch the CI setup to symfony flex, to have better control over the allowed symfony versions for a certain matrix

This commit https://github.com/nelmio/alice/pull/1284/commits/c764a2e9ffc784d30c9f7185551731b961bb887d should be reverted before merge as it is only needed to let the tests pass now while there is no stable release yet

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,6 +42,7 @@
         "symfony/serializer": "^6.4 || ^7.3 || ^8.0",
         "symfony/var-dumper": "^6.4 || ^7.3 || ^8.0"
     },
+    "minimum-stability": "beta",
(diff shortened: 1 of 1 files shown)


theofidry:
@acrobat Thanks a lot for the PR! I tried to update it in #1324 but there seems to be a bit more.

I spent most of the day updating a bit everything to unblock this and the PHP 8.5 upgrade, I need to go touch some grass. Maybe you can manage in the meantime, or I'll have another look during the week hopefully.

PR #1356, opened 2026-02-14
Add support for sebastian/comparator 8

stof:
This allows being compatible with projects using phpunit 13.

Note that I increased the phpunit range to actually have CI jobs installing those newer `sebastian/comparator` version (as each major version of PHPUnit is tied to a single major version of `sebastian/comparator`)

Closes #1355


stof:
thanks

